PR #260, opened 2015-01-20
Dsc 124

shuangzhou:
This request added ability to sync reference data between code impl and database.

### DIFF
--- a/src/amberdb/util/LookupRefresh.java
+++ b/src/amberdb/util/LookupRefresh.java
@@ -8,6 +8,8 @@ import javax.sql.DataSource;
 
 import org.h2.jdbcx.JdbcConnectionPool;
 import org.skife.jdbi.v2.DBI;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import amberdb.AmberSession;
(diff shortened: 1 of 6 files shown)


src/amberdb/util/LookupRefresh.java
```diff
@@ -0,0 +1,60 @@
...
+                // add new lookup entry
+                db.getLookups().addLookup(fromMap.get(nameCode));
+            } else {
+                // update an existing lookup entry
+                System.out.println("updating lookups for " + nameCode);
+                Long id = toMap.get(nameCode).getId();
+                db.getLookups().updateLookup(id, fromMap.get(nameCode).getValue());
+            }
+        }
+        db.commit();
```
scoen: this is a(n amber) session commit - not a database commit, so it is not necessary.

scoen: Actually, it should be removed. The caller might have made graph changes in the session and would not want this call to inadvertently commit them yet.

shuangzhou: Good idea, I'll make the change now.

PR #263, opened 2015-01-26
Dsc 1349

shuangzhou:
Hi Shaun,

This request includes updates to EAD metadata mappings requested in dsc-1349.  Could you please review these changes?

Thanks,
Shuang

### DIFF
--- a/src/amberdb/model/builder/ComponentBuilder.java
+++ b/src/amberdb/model/builder/ComponentBuilder.java
@@ -2,6 +2,7 @@ package amberdb.model.builder;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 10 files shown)


src/amberdb/model/builder/ComponentBuilder.java
```diff
@@ -103,28 +106,53 @@ protected static void mapWorkMD(EADWork componentWork, String uuid, Map<String,
...
+        else    
+            componentWork.asEADWork().setRdsAcknowledgementReceiver("NLA");
         componentWork.setEADUpdateReviewRequired("Y"); 
-        componentWork.setAccessConditions("Unrestricted");
         
-        Object componentLevel = fieldsMap.get("component-level");
-        if (componentLevel != null && !componentLevel.toString().isEmpty()) {
+        if (componentWork.getParent() == null)
+            System.out.println("component work " + componentWork.getObjId() + "'s parent is null.");
+        String accessConditions = componentWork.getParent().getAccessConditions();
```
scoen: This is going to error if parent is null.

src/amberdb/model/builder/ComponentBuilder.java
```diff
@@ -103,28 +106,53 @@ protected static void mapWorkMD(EADWork componentWork, String uuid, Map<String,
...
+        if (accessConditions != null && !accessConditions.isEmpty())
+            componentWork.setAccessConditions(accessConditions);
+        else
+            componentWork.setAccessConditions("Restricted");
+        componentWork.setDigitalStatus(DigitalStatus.NON_DIGITISED.code());
+        
+        List<String> constraints = componentWork.getParent().getConstraint();
```
scoen: This is going to error if parent is null also.

src/amberdb/model/builder/ComponentBuilder.java
```diff
@@ -103,28 +106,53 @@ protected static void mapWorkMD(EADWork componentWork, String uuid, Map<String,
...
+        componentWork.setConstraint(constraints);
+        
+        Date expiryDate = componentWork.getParent().getExpiryDate();
+        componentWork.setExpiryDate(expiryDate);
+        
+        String internalAccessConditions = "open";
+        if (componentWork.getCollection() != null && componentWork.getCollection().equalsIgnoreCase("nla.ms"))
+            internalAccessConditions = "restricted";
```
scoen: Just curious: Here (internalAccessConditions) 'restricted' is all lower case, but we use 'Restricted' for standard 'accessConditions'. On a broader level, maybe we need a constant or enum or something for both these sets of conditions (for access and internal access) ? Otherwise might need to always use equalsIgnoreCase. Just a thought.

src/amberdb/model/builder/ComponentBuilder.java
shuangzhou: I've checked that componentWork always have a parent in the EAD collection creation currently.  I've removed this check as the subsequent code will throw an error anyway if an componentWork doesn't have a parent unexpectedly.

scoen: Fair enough.

You could probably get rid of the System.out line then as it will never fire.

src/amberdb/model/builder/ComponentBuilder.java
shuangzhou: Good idea, I've added AccessCondition enum now.

PR #266, opened 2015-01-28
Move image representation selection logic into amberdb.

AaronCMuller:


AaronCMuller:
@terenceingram @JonathanShaw look please?

PR #268, opened 2015-01-29
Dsc 1351

shuangzhou:
This requests included additional EAD collection level fields required by delivery.


shuangzhou:
@scoen @bsingh-nla could you please review these changes?  I've also created nla.obj-8649844 which has works reflecting these changes.

Thanks,
Shuang

PR #269, opened 2015-02-02
Change repImage resolving to be simpler.

AaronCMuller:


AaronCMuller:
@JonathanShaw @terenceingram @bsingh-nla

scoen:
I’ll kick the amberdb version up so maven et al sees the changes

PR #270, opened 2015-02-02
change 'drop table' to 'drop temporary table'.

AaronCMuller:
This allows amberdb_ro to make amber queries.

@scoen


scoen:
It will break the tests – h2 doesn’t support it. Will need to include alternate code

PR #277, opened 2015-02-09
Dsc 1349

shuangzhou:
Hi Shaun,

This request includes new enumerations CopyType, CopyrightPolicy and MaterialType, and it is also updated to follow java convention of starting calendar month from 0 rather than 1.  Could you please review these changes?

Thanks,
Shuang

### DIFF
--- a/src/amberdb/enums/CopyrightPolicy.java
+++ b/src/amberdb/enums/CopyrightPolicy.java
@@ -1,0 +1,38 @@
+package amberdb.enums;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public enum CopyrightPolicy {
+    PERPETUAL("Perpetual"),
+    OUTOFCOPYRIGHT("Out of Copyright");
+    
+    private String code;
+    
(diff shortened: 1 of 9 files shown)


src/amberdb/enums/CopyrightPolicy.java
```diff
@@ -0,0 +1,38 @@
+package amberdb.enums;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public enum CopyrightPolicy {
```
terenceingram: Why is there an enum if the Copyright Policy is the in the lookups table?

shuangzhou: Currently CopyrightPolicy enum reference the lookup code (not the lookup display value) in the lookups table.  The lookup code can't be changed while the lookup display value can.  

I'm hoping that having CopyrightPolicy enum referenced in the app rather than a string can:
- enforce we reference lookup code rather than lookup display value; and
- can prevent the business logic in the app being unintentionally changed due to an update of an lookup entry.

PR #278, opened 2015-02-10
Bs deprecate fields first part

bsingh-nla:
dsc-388
@scoen, @shuangzhou , @JonathanShaw , @terenceingram can you please look.

### DIFF
--- a/src/amberdb/model/Work.java
+++ b/src/amberdb/model/Work.java
@@ -243,15 +243,19 @@ public interface Work extends Node {
     public void setCopyrightPolicy(String copyrightPolicy);
 
     @Property("firstPart")
+    @Deprecated
     public String getFirstPart();
 
     @Property("firstPart")
+    @Deprecated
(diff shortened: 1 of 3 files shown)


src/amberdb/model/Work.java
```diff
@@ -243,15 +243,19 @@
...
     @Property("firstPart")
+    @Deprecated
     public String getFirstPart();
 
     @Property("firstPart")
+    @Deprecated
     public void setFirstPart(String firstPart);
 
     @Property("sortIndex")
+    @Deprecated
```
shuangzhou: Why is sortIndex deprecated?  It's a read-only field referenced in dsc-1369.

bsingh-nla: According to task http://ourweb.nla.gov.au/apps/jira/browse/DSC-388,

Fields 'sortIndex' and 'firstPart' should be removed. Instead of deleting these methods, Terence suggested deprecating them.

PR #282, opened 2015-02-17
Make changes to allow multiple copies per work (where reasonable).

AaronCMuller:


AaronCMuller:
@scoen Could you take a look at this?

PR #285, opened 2015-02-19
Dscs 146

shuangzhou:
Hi Shaun,

This request includes fixes and unit tests for extracting correspondence index and container list from an EAD.
Could you please review these changes?

Thanks,
Shuang

### DIFF
--- a/src/amberdb/model/Node.java
+++ b/src/amberdb/model/Node.java
@@ -292,8 +292,8 @@ public interface Node extends VertexFrame {
             List<T> entries = new ArrayList<>();
             Iterator<Description> it = descriptions.iterator();
             while (it.hasNext()) {
-                Description next = it.next();
-                if (next.getType() != null && next.getType().equals(returnClass.getName())) {
+                Description next = it.next();  
+                if (next.getType() != null && (next.getType().equals(returnClass.getName()) || (next.getType().equals(returnClass.getSimpleName())))) {
                     entries.add(this.g().getVertex(next.asVertex().getId(), returnClass));
                 }
             }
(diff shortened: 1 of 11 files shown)


src/amberdb/model/Node.java
```diff
@@ -292,8 +292,8 @@ public Description getDescription(String fmt) {
             List<T> entries = new ArrayList<>();
             Iterator<Description> it = descriptions.iterator();
             while (it.hasNext()) {
-                Description next = it.next();
-                if (next.getType() != null && next.getType().equals(returnClass.getName())) {
+                Description next = it.next();  
+                if (next.getType() != null && (next.getType().equals(returnClass.getName()) || (next.getType().equals(returnClass.getSimpleName())))) {
```
scoen: I don't think the !=null is not protecting the 'or' (||) portion of this boolean statement.

I don't know if it needs to, but if getType() can return null there might be an issue

scoen: I don't think the !=null is protecting the 'or' (||) portion of this boolean statement. 

Is what i meant sorry

src/amberdb/model/builder/CollectionBuilder.java
```diff
@@ -606,65 +607,74 @@ protected static void processCollection(Work collectionWork, InputStream in, Jso
     
     protected static void extractFeatures(EADWork collectionWork, JsonNode featuresCfg, XmlDocumentParser parser) {
         String basePath = featuresCfg.get(CFG_BASE).getTextValue();
-        String repeatablePath = featuresCfg.get(CFG_REPEATABLE_ELEMENTS).getTextValue();
-        Map<String, String> mapping = parser.getFieldsMap(parser.doc.getRootElement(), featuresCfg, basePath);
+        Nodes nodes = parser.getElementsByXPath(parser.getDocument(), basePath);
+        if (nodes.size() > 0) {
```
scoen: could you indent the code in here ? Just formatting i know.

src/amberdb/model/Node.java
scoen: ah i see what you mean now - carry on nothing to see here ...

src/amberdb/model/builder/CollectionBuilder.java
shuangzhou: it's been reformatted.

PR #289, opened 2015-02-25
Dscs 127

shuangzhou:
Hi Shaun,

This request adds a check for existing work with an archival space id during reload of an EAD, and mark it for EAD update review.  Could you please review these changes?

Thanks,
Shuang

### DIFF
--- a/src/amberdb/model/builder/CollectionBuilder.java
+++ b/src/amberdb/model/builder/CollectionBuilder.java
@@ -27,8 +27,11 @@ import org.codehaus.jackson.node.ObjectNode;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.tinkerpop.blueprints.Vertex;
+
 import doss.core.Writables;
 
+import amberdb.NoSuchObjectException;
 import amberdb.PIUtil;
 import amberdb.enums.AccessCondition;
 import amberdb.enums.CopyRole;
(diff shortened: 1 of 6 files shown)


src/amberdb/model/builder/CollectionBuilder.java
```diff
@@ -891,14 +891,21 @@ private static void mapBibliography(Work collectionWork, Map<String, String> fie
...
+    protected static EADWork mapWorkMD(EADWork collectionWork, Node eadElement, JsonNode elementCfg, XmlDocumentParser parser) throws EADValidationException, JsonParseException, JsonMappingException, IOException {
+        EADWork workInCollection = null;
         Map<String, String> fieldsMap = parser.getFieldsMap(eadElement, elementCfg, parser.getBasePath(parser.getDocument()));        
-        if (fieldsMap.get("uuid") == null || fieldsMap.get("uuid").isEmpty())
+        if (fieldsMap.get("uuid") == null || fieldsMap.get("uuid").isEmpty()) {
             throw new EADValidationException("Failed to process collection " + parser.collectionObjId + " as no Archive Space id found for component work " + workInCollection.getObjId());
-        
+        }
         String uuid = fieldsMap.get("uuid");
+        workInCollection = collectionWork.checkEADWorkInCollectionByLocalSystemNumber(uuid);
```
scoen: just a thought - instead of returning null from this method, why not return addEADWork() and avoid the check below ?

PR #290, opened 2015-02-25
retaining access restrictions and capture status after ead reupload.

shuangzhou:
Hi Shaun,

This fix caters for the request of retaining access restrictions and capture status after ead reupload.
Could you please review this change?

Thanks,
Shuang

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 AmberDB                   
 =======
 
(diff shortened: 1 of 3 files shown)


scoen:
Hi Snezana,

As we talked about � can you just test that :id = List<Long> thing.

If we might need to fix that first if it is an issue.

Otherwise looks fine

shaun

PR #297, opened 2015-03-09
Dscs 125

shuangzhou:
This request includes fix for container extraction and data validation points.


snezanam:
Just wandering  about 
if (eadEntityEntries.size() == 0) {
String ord = (i == 0)? "1st" : (i == 1)? "2nd" : (i == 2)? "3rd" : "" + i + "th";
throw new EADValidationException("FAILED_EXTRACT_ENTITIES", ord, collectionWork.getObjId());
}

if ord should be  i + 1 in all cases or just  when i is  0-2

shuangzhou:
Yep, well spotted, done.

scoen:
Unsure how many entities are expected, but it will write funny things after the 20th eg: 21th, 22th and 23th etc. Still, i expect this will be rare.

shuangzhou:
Good point, I haven't thought through this very well, will try to come up with a better solution.

PR #305, opened 2015-03-26
Sc fix cast

scoen:
Hi @JonathanShaw could you take a look ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 AmberDB                        
 =======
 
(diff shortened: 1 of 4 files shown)


JonathanShaw:
nice, looks good to me

JonathanShaw:
@scoen I'm not seeing this in the repo.  do we need to kick jenkins/push a new version?

scoen:
Nexus sucks. It�s likely there, just not indexed yet. Probably have to get nexus to do a re-index

PR #309, opened 2015-03-31
updated listUUIDs() for ead reload.

shuangzhou:
listUUIDs() is expected to extract all UUIDs from an EAD ingest file for comparison with existing EAD works within a collection.  Currently, it's only extracting the first level of children from the file, and not traversing to all levels within the file.

This request provides the fix to extract all UUIDs from an EAD ingest file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 AmberDB                        
 =======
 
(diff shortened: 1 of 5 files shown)


shuangzhou:

PR #310, opened 2015-04-01
Dscs 266

shuangzhou:
This request provides ability to extract paragraphs from feature <odd> elements (as well as container list) at ead collection level in ead ingest file.


shuangzhou:

PR #313, opened 2015-04-09
Acm/dscs-121 Add additional properties to sections

AaronCMuller:
@terenceingram @bsingh-nla Please take a look.

### DIFF
--- a/src/amberdb/model/Section.java
+++ b/src/amberdb/model/Section.java
@@ -44,6 +44,30 @@ public interface Section extends Work {
 	
 	@JavaHandler
 	public int countExistsOns();
+
+	@Property("abstract")
+	public String getAbstract();
+
+	@Property("abstract")
+	public void setAbstract(String abstractText);
+
+	@Property("captions")
+	public List<String> getCaptions();
+
+	@Property("captions")
+	public void setCaptions(List<String> captions);
+
+	@Property("advertising")
+	public boolean hasAdvertising();
+
+	@Property("advertising")
+	public void setAdvertising(boolean advertising);
+
+	@Property("illustrated")
+	public boolean isIllustrated();
+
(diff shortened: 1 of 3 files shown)


src/amberdb/model/Section.java
```diff
@@ -44,6 +44,36 @@
 	
 	@JavaHandler
 	public int countExistsOns();
+
+	@Property("abstract")
+	public String getAbstract();
+
+	@Property("abstract")
+	public void setAbstract(String abstractText);
```
shuangzhou: @AaronCMuller , this field already exist in Work, Section automatically inherit this.

src/amberdb/model/Section.java
```diff
@@ -44,6 +44,36 @@
...
+
+	@Property("captions")
+	public List<String> getCaptions();
+
+	@Property("captions")
+	public void setCaptions(List<String> captions);
+
+	@Property("advertising")
+	public boolean isAdvertising();
```
shuangzhou: @AaronCMuller could you please change return type to Boolean?

src/amberdb/model/Section.java
```diff
@@ -44,6 +44,36 @@
...
+
+	@Property("advertising")
+	public void setAdvertising(boolean advertising);
+
+	@Property("illustrated")
+	public boolean isIllustrated();
```
shuangzhou: @AaronCMuller could you please change return type to Boolean?

AaronCMuller: @shuangzhou Can you tell me what difference it makes?

shuangzhou: @AaronCMuller the field might not be always set, from memory, this will return null instead of a boolean value.

PR #314, opened 2015-04-12
some code refactor for collection builder.

shuangzhou:
This request refactored some reference to List to Set for better implementation and faster access.


shuangzhou:

PR #322, opened 2015-04-27
Update comments for common cases.

AaronCMuller:
@terenceingram @bsingh-nla @ninhnguyen @shuangzhou

### DIFF
--- a/src/amberdb/model/Node.java
+++ b/src/amberdb/model/Node.java
@@ -148,11 +148,17 @@ public interface Node extends VertexFrame {
     @Property("commentsInternal")
     public void setCommentsInternal(String commentsInternal);
 
+    @JavaHandler
+    public void addCommentsInternal(String comment);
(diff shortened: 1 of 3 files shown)


src/amberdb/model/Node.java
```diff
@@ -314,6 +320,31 @@ public void setAlias(List<String> alias) throws JsonParseException, JsonMappingE
...
+                setCommentsInternal(commentsInternal + "\n" + comment);
+            }
+            else {
+                setCommentsInternal(comment);
+            }
+        }
+
+        @Override
+        public void addCommentsExternal(String comment) {
+            String commentsExternal = getCommentsExternal();
```
shuangzhou: be good to have the impl for both addCommentsInternal and addCommentsExternal in one private method, as addCommentsExternal should check for empty or null string condition as well.

PR #324, opened 2015-04-27
add check condition for potential invalid date with number or hyphen.

shuangzhou:
This request further improve on the date parser's ability to recognize invalid date expression strings during date range extraction.  It now filters out strings containing only 1-2 digits number that is neither year, or month of any particular year or date in any particular month of any particular year.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 AmberDB                       
 =======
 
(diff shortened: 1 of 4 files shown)


shuangzhou:

PR #325, opened 2015-04-28
updated find model apis.

shuangzhou:
This request attempt to provide inter-operability between amber vertex and tinkerpop wrapped vertex.


shuangzhou:
There's still additional conditions needs to be covered, will investigate more.

PR #327, opened 2015-04-29
updated to find vertex in framed graph.

shuangzhou:
The request update find api to find vertices in framed graph.


shuangzhou:

PR #331, opened 2015-05-07
Remove JSON generation as it's not required by business or us.

AaronCMuller:
Note that I considered making CollectionBuilder a non-static class. Then there would be the potential to lazy-load the collection works into a field, which has potential time savings if there are multiple calls to getObjsInTree. Thoughts?

@scoen @shuangzhou @terenceingram @bsingh-nla  Please look

### DIFF
--- a/src/amberdb/model/builder/CollectionBuilder.java
+++ b/src/amberdb/model/builder/CollectionBuilder.java
@@ -1,80 +1,36 @@
 package amberdb.model.builder;
 
-import java.io.IOException;
-import java.io.InputStream;
-import java.text.ParseException;
-import java.text.SimpleDateFormat;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.Date;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.List;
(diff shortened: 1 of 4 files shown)


src/amberdb/model/builder/CollectionBuilder.java
```diff
@@ -938,154 +852,5 @@ protected static EADWork mapWorkMD(EADWork collectionWork, Node eadElement, Json
         ComponentBuilder.mapWorkMD(workInCollection, uuid, fieldsMap);
         return workInCollection;
```
shuangzhou: might be worthwhile catching a general exception and wrap this in the EADValidationException and pass in the archive spaces id in the EADValidationException.  My bad, I meant to do this before, just hasn't got around to it.

shuangzhou:
re: I considered making CollectionBuilder a non-static class

I thought about this before as well, was just concerned about it might take up more memory that way, maybe Document class can be refactored to achieve what you were after, not sure.

shuangzhou:
EAD Json is referenced by the EAD job processing in banjo currently to obtain a list of EAD works requiring review, this'll need to be updated to work with this change.

shuangzhou:
Thanks for refactoring the code.

PR #332, opened 2015-05-19
Dscs 267

shuangzhou:
This request provides update on the date parsing rule to look at the normal attribute in the unitdate element first.  It also include several general improvements on exceptions and reading EAD file with utf8 support.


shuangzhou:
@scoen @ninhnguyen could you please review these changes?  Thanks.

src/amberdb/model/builder/ComponentBuilder.java
```diff
@@ -108,138 +108,160 @@ public static EADWork mergeComponent(EADWork collectionWork, EADWork parentWork,
...
+            log.debug("work " + work.getObjId() + ": date range: " + dateRange);
+            work.setDateRangeInAS(dateRange);
             List<Date> dateList;
             try {
-                dateList = DateParser.parseDateRange(dateRange.toString());
+                dateList = DateParser.parseDateRange(dateRange);
                 if (dateList != null && dateList.size() > 0) {
-                    if (dateList.get(0) == null) {
-                        throw new EADValidationException("FAILED_TO_DETERMINE_START_DATE", componentWork.getObjId(), (uuid == null)?"":uuid);
+                    if (DateParser.isToday(dateList.get(0))) {
```
scoen: Why do you throw an exception if the range includes today's date as one of the end points of the range ? Isn't this risky, as i expect it is possible for a range to have this date ?

shuangzhou:
Ok, code is updated, just waiting for the travis to build.

PR #334, opened 2015-05-21
Dscs 361

shuangzhou:
This request provides additional api for retrieve a list of copies for a work in an orderly fashion as per requirements in dscs-361.


shuangzhou:
@scoen could you please review this change?  Thanks.

PR #335, opened 2015-05-28
changed getter and setter for constraints to return sets.

shuangzhou:
This request provides update to constraints api as constraints selected need to be in a unique set as per dscs-227.

### DIFF
--- a/src/amberdb/model/Work.java
+++ b/src/amberdb/model/Work.java
@@ -8,12 +8,12 @@ import java.util.Collection;
 import java.util.Date;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
 
 import amberdb.util.WorkUtils;
 import com.google.common.collect.Iterables;
-import com.google.common.collect.Sets;
 import org.codehaus.jackson.JsonParseException;
 import org.codehaus.jackson.map.JsonMappingException;
 import org.codehaus.jackson.map.ObjectMapper;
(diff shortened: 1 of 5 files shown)


shuangzhou:
@scoen could you please review this change?  Thanks.

terenceingram:
As lists are ordered and the sets are not ordered, I wonder if that will make a difference somewhere in the UI or somewhere else?

shuangzhou:
Good point, the calling app should use the LinkedHashSet to construct the multi-valued constraint.  I'll update the getConstraint to retain order.

terenceingram:
I wonder how freemarker handles Sets? I wonder if you access it the same way as a list in freemarker.

shuangzhou:
Looks like you can access it the same in freemarker.

src/amberdb/model/Work.java
```diff
@@ -1270,12 +1272,17 @@ public IsPartOf getParentEdge() {
...
 
         @Override
-        public List<String> getConstraint() throws JsonParseException, JsonMappingException, IOException {          
-            return deserialiseJSONString(getJSONConstraint());
+        public Set<String> getConstraint() throws JsonParseException, JsonMappingException, IOException { 
+            List<String> list = deserialiseJSONString(getJSONConstraint());
+            LinkedHashSet<String> constraint = new LinkedHashSet<>();
+            for (String entry : list) {
+                constraint.add(entry);
+            }
```
scoen: can you use addAll here to reduce the lines of code ?

shuangzhou:
It's updated.

terenceingram:
@shuangzhou @scoen Shuang, I forgot to add that you may need to test the indexing code in Jelly. As you have changed the return type, I am not sure what the code that is indexing is expecting when it gets the constraint. It would definitely iterate over the list to add it as a multivalued field.

shuangzhou:
@terenceingram I'll update Jelly with this change.  Thanks for the suggestion.

PR #341, opened 2015-06-11
dscs-418 - Detect duplicate IDs in EADs

AaronCMuller:
@JonathanShaw @terenceingram @ninhnguyen @bsingh-nla @scoen


scoen:
Looks fine to me, but i ain't no EAD expert. Anyone else ?

PR #342, opened 2015-06-15
Dscs 515

shuangzhou:
This request includes general code tidy up for items from findbugs including the slight code refactor in the VersionedVertex.

@scoen could you please review these changes?  Thanks.

### DIFF
--- a/src/amberdb/graph/AmberHistory.java
+++ b/src/amberdb/graph/AmberHistory.java
@@ -97,14 +97,14 @@ public class AmberHistory {
         for (VersionedVertex v : vertices) {
             TVertexDiff diff = v.getDiff(txn1, txn2);
             TTransition change = diff.getTransition();
-            TId id = diff.getId()[0];
+            Long id = diff.getId()[0].getId();
 
             // Deletion trumps all changes - don't replace
             if (modifiedIds.get(id) != null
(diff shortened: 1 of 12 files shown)


src/amberdb/graph/AmberHistory.java
```diff
@@ -97,14 +97,14 @@ public VersionedGraph loadChangedGraphForPeriod(Long txn1, Long txn2) {
         for (VersionedVertex v : vertices) {
             TVertexDiff diff = v.getDiff(txn1, txn2);
             TTransition change = diff.getTransition();
-            TId id = diff.getId()[0];
+            Long id = diff.getId()[0].getId();
```
scoen: this isn't going to work. see the deletion if-continue bit below

scoen: apologies - this will certainly work and is a definite improvement  - may even fix a bug i've  seen - nice on Shuang
my mistake

src/amberdb/version/VersionedVertex.java
```diff
@@ -170,6 +170,9 @@ public String toString() {
         return vertices;
     }
     
+    public int hashCode() {
+        return super.hashCode();
```
shuangzhou: @scoen, thanks for spotting this and suggest the change.  I've updated the hashcode() to return the value from the object.

PR #343, opened 2015-06-16
Bs dscs 429 marc fields

bsingh-nla:
@scoen , @shuangzhou  can you please look.


bsingh-nla:
@scoen can you please look again.

PR #344, opened 2015-06-18
Dscs 434

shuangzhou:
This request added ocr check methods that can be called during and post ocr json generation to verify whether the ocr bounding coords are valid.


scoen:
just checking i read this correctly - if there is no image then this check returns false (ie: not out of bounds) ?

shuangzhou:
thanks for spotting this.  no, it's not correct, I'll update the code.

shuangzhou:
updated.

PR #346, opened 2015-06-19
Add holding id as a property

AaronCMuller:


AaronCMuller:
@terenceingram @JonathanShaw @scoen @yetti @bsingh-nla

PR #349, opened 2015-06-24
Dscs 410

shuangzhou:
Slight update to the ocr check impl to check for zoned text in zs in absence of ps, and made public the api to check ocr for ocr json from an inputstream.


shuangzhou:
@scoen could you please review this change?

PR #351, opened 2015-06-24
Dscs 457

snezanam:
Added the method to return Algorithm lookup based on carrier Name from carrier_algorithms, @shuangzhou  could you please have a look


shuangzhou:
looks good.

PR #354, opened 2015-06-30
changes for spanning interviews

yetti:
@terenceingram @JonathanShaw @AaronCMuller @scoen @shuangzhou @snezanam @ninhnguyen please have a look... Not really sure about the relation stuff between the parent delivery work (interview) and the children delivery works. Figured the media on an interview "exists on" 1 or more delivery works. Happy to change if it doesn't makes sense and anyone has an alternative.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,7 @@
 
   <groupId>au.gov.nla</groupId>
   <artifactId>amberdb</artifactId>
-  <version>1.1.325-SNAPSHOT</version>
+  <version>1.1.326-SNAPSHOT</version>
   <packaging>jar</packaging>
   <name>amberdb</name>
   <description>Digital library domain model</description>
--- a/src/amberdb/enums/SubType.java
+++ b/src/amberdb/enums/SubType.java
@@ -12,7 +12,8 @@ public enum SubType {
     VOLUME("volume"),
     COLLECTION("collection"),
     SERIES("series"),
-    FOLDER("folder");
+    FOLDER("folder"),
+    DELIVERY_WORK("delivery work");
 
     private String code;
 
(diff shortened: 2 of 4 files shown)


JonathanShaw:
Looks reasonable to me,  I'm not sure we need the new SubType? works of that subtype will all also have the setMaterialFromMultipleSources property set wouldn't they?  and no non delivery work would set the setMaterialFromMultipleSources property... I think

I haven't looked at the banjo side implementation so I could well be wrong here.

yetti:
Yeah, I had added the SubType after seeing the getExistsOn() method required a SubType as a param, but I don't actually use it in banjo. I was going to remove this after speaking with Aaron a while back, but must've missed it.

terenceingram:
I would prefer not to use subtype. We have tried to minimise it's use.

yetti:
Yep, it's been removed.

scoen:
Are the remove methods supposed to delete the interview or deliveryWork being removed ?

If so, there�d be an unfortunate problem:

Tinkerpop annotated �remove� methods just remove the connection between 2 nodes without actually removing the �removed� node (it removes the edge that connects the nodes).

I don�t know if this is what you wanted (maybe it is), but if not, then we�ll end up with orphaned works whenever this method is called.

You might want to see the delete methods in AmberSession if this isn�t the desired functionality.


shaun

yetti:
All the delivery works are meant to be existing Works (i.e. auto-ingested). All I'm doing here is creating relationships between the interview and the delivery works. As such, I don't think we'd want to delete forever the delivery work if we remove the relationship, but I'll clarify this with Kavinga.

scoen:
Cheers Yetrina,

I just wanted to check because we�ve done it before (left a whole bunch of orphaned Copies hanging around :-)

shaun

PR #358, opened 2015-07-06
Added methods to check if a work has born digital / master copy

ninhnguyen:
@scoen @JonathanShaw please have a look


scoen:
i think that looks fine.

PR #359, opened 2015-07-06
fix bug that saves the string 'null' when serialising to json a null …

scoen:
…value

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 AmberDB                       
 =======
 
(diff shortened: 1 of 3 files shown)


scoen:
@JonathanShaw @AaronCMuller @bsingh-nla @yetti @snezanam @sjacob 
Would someone take a look at this change ? I need to get it in to fix a rather significant bug that is causing problems with indexing. It will require a data fix also, but that's another story.

PR #360, opened 2015-07-21
dscs 867: Some refactoring, and also trim values so they do not contain newlines, etc

AaronCMuller:
@scoen @yetti @terenceingram @JonathanShaw


scoen:
Yep - looks good

PR #362, opened 2015-07-23
ms445

MingtaoSun:
@ninhnguyen move the logic of converting duration format from banjo to here


MingtaoSun:
@ninhnguyen @shuangzhou updated readme

PR #369, opened 2015-07-30
Sc big commit fix

scoen:


MingtaoSun:
Seems very sophisticated to me :+1:

scoen:
Yeah � it�s not nice.
I�m considering removing some of it now. I think I should do at least 2 separate pull requests. It is addressing 2 separate concerns.
don�t merge yet

PR #371, opened 2015-07-30
Sc process hang fix

scoen:


src/amberdb/model/Copy.java
```diff
@@ -165,6 +163,18 @@
...
+
+    @Property("acquisitionStatus")
+    public void setAcquisitionStatus(String acquisitionStatus);
+
+    @Property("acquisitionCategory")
+    public String getAcquisitionCategory();
+
+    @Property("acquisitionCategory")
+    public void setAcquisitionCategory(String acquisitionCategory);
+
```
MingtaoSun: Oh, my stuff is also there. I just learned to prevent this, I need to do a pull origin master on my 'master' branch before creating a working branch.

PR #375, opened 2015-08-05
added more details for the work test and check for generating lookup code.

shuangzhou:
This request is part of dscs-1071 to add more unit tests.

### DIFF
--- a/src/amberdb/sql/Lookups.java
+++ b/src/amberdb/sql/Lookups.java
@@ -155,6 +155,10 @@ public abstract class Lookups extends Tools {
    
     
     public Long addLookup(String name, String code, String value) {
+        List<ListLu> list = findActiveLookup(name, code);
+        if (!list.isEmpty()) {
+            code = code + "_" + list.size();
+        }
         System.out.println("lookups: start time : " + fmtDate(new Date()));
         Long newId = addLookupData(name, code, value);
         System.out.println("lookups: end time : " + fmtDate(new Date()));
--- a/src/amberdb/sql/Lookups.java
+++ b/src/amberdb/sql/Lookups.java
@@ -153,7 +153,18 @@ public abstract class Lookups extends Tools {
     }
     
    
-    
+    /**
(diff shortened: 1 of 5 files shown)


src/amberdb/sql/Lookups.java
```diff
@@ -155,6 +155,10 @@ public CarrierAlgorithm map(int index, ResultSet r, StatementContext ctx) throws
    
     
     public Long addLookup(String name, String code, String value) {
+        List<ListLu> list = findActiveLookup(name, code);
+        if (!list.isEmpty()) {
+            code = code + "_" + list.size();
+        }
```
scoen: i'm unsure this is desired behaviour for two reasons : 
1. Is it possible that you might actually want multiple values to map to the same code ? This change would block that occurring. I'm assuming value is what the user sees and code is what the app works on. 
2. The user of the function has asked for a particular code to be added, but internally a different one has been added. This is risky behaviour.

So i guess i'm asking what use cases this extra functionality has been added to support?

shuangzhou: Good questions.
re: 1, the reason for this change is because the code is currently being referenced in amberdb and calling apps, so it's nice to ensure its unique.  Currently there's no requirement to map multiple value to the same code.

re: 2, code is also meant to be system generated, transparent to people using the app.  At the moment, by default, the code will have the value supplied when adding a new entry, it looks like an central and reasonable place to cater for this.

MingtaoSun: What problem have we encountered with the old addLookup() method?

shuangzhou: This is an improvement, currently it is assumed that the calling method will ensure that the code input is unique and has not already being referenced anywhere.

scoen:
Then do you need to do the same for the other addLookup method ?

scoen:
re: 1. You also haven't prevented the same value pointing to multiple codes and that is where the real issue could occur: Two options with the same displayed value leads to 2 possible (invisible) codes being sent to the app ?

PR #376, opened 2015-08-06
ms1051: add new field to Work

MingtaoSun:
@scoen @shuangzhou

### DIFF
--- a/src/amberdb/model/Work.java
+++ b/src/amberdb/model/Work.java
@@ -165,6 +165,12 @@ public interface Work extends Node {
 
     @Property("form")
     public void setForm(String form);
+    
+    @Property("displayTitlePage")
+    public String getDisplayTitlePage();
+
+    @Property("displayTitlePage")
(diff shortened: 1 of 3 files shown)


src/amberdb/model/Work.java
```diff
@@ -165,6 +165,12 @@
 
     @Property("form")
     public void setForm(String form);
+    
+    @Property("displayTitlePage")
+    public String getDisplayTitlePage();
```
scoen: Because this is a flag, perhaps it might be better as a boolean property ?

MingtaoSun:
@scoen, yep, it needs to be a checkbox, so I have changed it to boolean. Thanks.

AaronCMuller:
There is really no reason to update the snapshot version here. Snapshots are meant to be overriden.

scoen:
Aaron is right. 
Let's try not updating the SNAPSHOT version every time.

MingtaoSun:
@AaronCMuller @scoen Fixed

PR #377, opened 2015-08-06
ms965: add new copy role

MingtaoSun:
@scoen @terenceingram

### DIFF
--- a/src/amberdb/enums/CopyRole.java
+++ b/src/amberdb/enums/CopyRole.java
@@ -58,7 +58,9 @@ public enum CopyRole {
     THUMBNAIL_COPY("t", "Thumbnail",43),
     TIME_CODED_SUMMARY("sc", "Time coded Summary", "Yes",44),
     TIME_CODED_TRANSCRIPT_COPY("tc", "Time coded transcript", "Yes",45),
-    VIEW_COPY("v", "View",46);
+    VIEW_COPY("v", "View",46),
+    Image_Package("ip", "Image Package",46);
+	
 
     private String code;
     private String display;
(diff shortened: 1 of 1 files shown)


src/amberdb/enums/CopyRole.java
```diff
@@ -58,7 +58,9 @@
     THUMBNAIL_COPY("t", "Thumbnail",43),
     TIME_CODED_SUMMARY("sc", "Time coded Summary", "Yes",44),
     TIME_CODED_TRANSCRIPT_COPY("tc", "Time coded transcript", "Yes",45),
-    VIEW_COPY("v", "View",46);
+    VIEW_COPY("v", "View",46),
+    IMAGE_PACKAGE("ip", "Image Package",47);
```
terenceingram: Just wondering what the numbers are? 46, 47 etc. Is that the id in the database table? If that is the case then this enum is just horrible.

scoen: agreed. I suspect it was added as an ordering mechanism by someone :(

terenceingram: Very sad. Might add it to the list to fix. Pity we didn't pick it up in a code review :(

MingtaoSun: for ordering purpose. I don't think we store copy roles as reference data in the database.

scoen: Yeah, i think the problem with this approach is that everything after the first 4 or so entries everything else is alphabetical, so to add IMAGE_PACKAGE entry in the correct place (alphabetically) you'd have to renumber everything after it which would be a gigantic pita. So Image package will appear at the end of the list now. Sad face.

AaronCMuller: This is an ordering mechanism to ensure copy roles are shown in the correct order.

You can possibly get around this by externalising the order to a Map or paired list of some kind, but even if you do that you still have the same problem if you want to insert something with a similar order in the middle.

If it is indeed only the first four things that matter, then yes ... this should probably be refactored.

PR #378, opened 2015-08-11
Ms229

MingtaoSun:
@scoen @terenceingram @AaronCMuller @bsingh-nla 
http://ourweb.nla.gov.au/apps/jira/browse/DSCS-229
Need to get the parent PI history


src/amberdb/version/VersionedGraph.java
```diff
@@ -153,12 +153,27 @@ protected void addVertexToGraph(VersionedVertex v) {
     
     
     public VersionedEdge getEdge(Long id) {
-        return getEdge(id, localMode);
+        String sql = "SELECT id, txn_start, txn_end, label, v_in, v_out, edge_order "
```
AaronCMuller: does this not do the same thing as getEdge?

scoen:
Ming i don't understand - why are you modifying this ?

src/amberdb/version/VersionedGraph.java
AaronCMuller: Oh I see, you've changed the params of the other getEdge

MingtaoSun: I have created another method getEdgeByLabelAndOutVertexId(), so I refactored the existing one for code reusing purpose

src/amberdb/version/VersionedGraph.java
```diff
@@ -153,12 +153,27 @@ protected void addVertexToGraph(VersionedVertex v) {
...
+    }
+    
+    public VersionedEdge getEdgeByLabelAndInVertexId(Long inVertexId, String label){
+        String sql = "SELECT id, txn_start, txn_end, label, v_in, v_out, edge_order "
+                + " FROM edge " 
+                + " WHERE v_in = :v_in_id and label = :label";
+        return getEdge(inVertexId, localMode, sql, new Object[]{inVertexId, label});
+    }
+    
+    protected VersionedEdge getEdge(Long eId, boolean localOnly, String sql, Object[] params) {
```
AaronCMuller: can params not be varargs?

MingtaoSun:
@scoen , I created another method getEdgeByLabelAndOutVertexId(). I modified the existing one to reuse the code.

AaronCMuller:
This isn't actually required at all. The functionality that you're after is already provided by
VersionedVertex#getEdges

scoen:
I think it'd be better to use the VersionedVertex if possible

PR #379, opened 2015-08-11
ms229: add utility method

MingtaoSun:
@scoen @AaronCMuller 
http://ourweb.nla.gov.au/apps/jira/browse/DSCS-229
Just added a utility method to find the correct TEdge


src/amberdb/version/VersionedEdge.java
```diff
@@ -110,4 +112,22 @@ public TEdgeDiff getDiff(Long txn1, Long txn2) {
     public List<TEdge> getTEdges() {
         return edges;
     }
+    
+    public static TEdge findEdgeByTransactionId(Iterable<VersionedEdge> versionedEdges, Long transactionId) {
```
scoen: More than 1 Edge may be live at any given transaction. This method only finds the first in the iteration. Perhaps rename as findFirstEdge...

MingtaoSun: Good idea! :+1:

src/amberdb/version/VersionedEdge.java
```diff
@@ -110,4 +112,22 @@ public TEdgeDiff getDiff(Long txn1, Long txn2) {
     public List<TEdge> getTEdges() {
         return edges;
     }
+    
+    public static TEdge findEdgeByTransactionId(Iterable<VersionedEdge> versionedEdges, Long transactionId) {
+        for (VersionedEdge versionedEdge : versionedEdges){
+            TEdge edge = versionedEdge.getAtTxn(transactionId);
+            if (edge != null){
+                edge.replaceProperties(getEdgeProperties(edge));
```
scoen: I don't understand whats happening here

MingtaoSun: You mean calling replaceProperties() method? We need to specify which properties to compare. Otherwise calling getDiff() method returns nothing or generate error if properties is null.

MingtaoSun:
@scoen , please have a look. I have tested, seems working fine. Thanks.

src/amberdb/version/TEdgeDiff.java
```diff
@@ -69,6 +71,31 @@ public Object getProperty(String propertyName) throws TDiffException {
...
+                diffMap.put(key, new Object[] {null, values[0]});
+                break;
+            case DELETED:
+                diffMap.put(key, new Object[] {values[0], null});
+                break;
+            case MODIFIED:
+                if (values.length == 2){
+                    diffMap.put(key, values);    
+                }else if(values.length == 1){
+                    diffMap.put(key, new Object[] {values[0], values[0]});
```
scoen: should this be values[0], values[1] ?

src/amberdb/version/TEdgeDiff.java
```diff
@@ -69,6 +71,31 @@ public Object getProperty(String propertyName) throws TDiffException {
         return super.getProperty(propertyName);
     }
     
+    public Map<String, Object[]> getDiffMap() {
+        Map<String, Object[]> diffMap = new HashMap<>();
```
scoen: instead use:
`Map<String, Object[]> diffMap = super.getDiffMap();`
As this will include the edges properties if they ever get any

scoen:
Apart from other comments looks good

src/amberdb/version/TEdgeDiff.java
MingtaoSun: length = 1, so value[1] would be out of boundary

scoen: no, in the case of MODIFIED you'll have a value for the first edge and a value for the 2nd edge

scoen: sorry, i didn't read it properly - my bad

MingtaoSun: I think sometimes 1 object sometimes 2 objects?

scoen: yes you're right

src/amberdb/version/TElement.java
```diff
@@ -46,7 +46,9 @@ public TId getId() {
     
     
     protected void replaceProperties(Map<String, Object> newProperties) {
-        properties = newProperties;
+        if (newProperties != null){
```
MingtaoSun: @scoen , need to add this check, Currently it's broken, because at somewhere, null was passed in

MingtaoSun:
@scoen , called the super.getDiffMap()

PR #384, opened 2015-08-18
DSCS-269 Improve error handling on EAD uploads

danielbaker:
@JonathanShaw @AaronCMuller @MingtaoSun 

Improved error handling on EAD uploads
 - fixed several error messages which didnt work
 - Added UUID to some error messages
 - Fixed an issue where bad UTF-8 Chars broke the xml parsing

### DIFF
--- a/src/amberdb/model/builder/CollectionBuilder.java
+++ b/src/amberdb/model/builder/CollectionBuilder.java
@@ -10,6 +10,7 @@ import nu.xom.Node;
 import nu.xom.Nodes;
 import nu.xom.ParsingException;
 import nu.xom.ValidityException;
+import org.apache.commons.io.input.BOMInputStream;
 import org.apache.commons.lang.StringUtils;
 import org.codehaus.jackson.JsonNode;
 import org.codehaus.jackson.JsonParseException;
@@ -90,7 +91,7 @@ public class CollectionBuilder {
         }
         
         // initializing the parser
-        try (InputStream fileStream = eadFile.openStream()) {
+        try (InputStream fileStream = new BOMInputStream(eadFile.openStream(), false)) {
             parser.init(collectionWork.getObjId(), fileStream, collectionCfg);
         }
         processCollection(collectionWork, collectionCfg, parser);
@@ -330,7 +331,7 @@ public class CollectionBuilder {
         Set<String> currentDOs = digitisedItemList(collectionWork);
         
         // initializing the parser
-        parser.init(collectionWork.getObjId(), eadFile.openStream(), collectionCfg);
+        parser.init(collectionWork.getObjId(), new BOMInputStream(eadFile.openStream(), false), collectionCfg);
         
         // Step 1: process collection from EAD: 
         //          - compare and update the metadata in collectionWork from the updated EAD finding aid header.
@@ -370,6 +371,9 @@ public class CollectionBuilder {
 
         Set<String> dupUuids = parser.listDuplicateUUIDs();
         if (dupUuids.size() != 0) {
+            if (dupUuids.contains("")) {
+                throw new EADValidationException("MISSING_UUID_DETECTED", "");
+            }
             throw new EADValidationException("DUPLICATE_UUID", Joiner.on(", ").join(dupUuids));
         }
 
--- a/src/amberdb/model/builder/CollectionBuilder.java
+++ b/src/amberdb/model/builder/CollectionBuilder.java
@@ -369,13 +369,7 @@ public class CollectionBuilder {
         boolean newCollection = true;
         Map<String, String> componentWorks;
 
-        Set<String> dupUuids = parser.listDuplicateUUIDs();
-        if (dupUuids.size() != 0) {
-            if (dupUuids.contains("")) {
-                throw new EADValidationException("MISSING_UUID_DETECTED", "");
-            }
-            throw new EADValidationException("DUPLICATE_UUID", Joiner.on(", ").join(dupUuids));
-        }
+        new EADDuplicateValidator(parser).validate();
 
         if (collectionWork.getChildren() != null && collectionWork.getChildren().iterator().hasNext()) {
             newCollection = false;
--- a/src/amberdb/model/builder/ComponentBuilder.java
+++ b/src/amberdb/model/builder/ComponentBuilder.java
@@ -148,9 +148,9 @@ public class ComponentBuilder {
                         + componentLevel.toString());
 
                 SubUnitType subUnitType = SubUnitType.fromString(componentLevel.toString());
-                if (subUnitType == null)
-                    throw new IllegalArgumentException("Invalid subunit type " + componentLevel.toString()
-                            + " for work " + componentWork.getObjId() + ".");
+                if (subUnitType == null) {
+                	throw new EADValidationException("INVALID_SUB_UNIT_TYPE", componentLevel.toString(), (uuid == null) ? "" : uuid);
+                }
                 componentWork.setSubUnitType(subUnitType.code());
                 // determine bib level with business rule borrowed from DCM
                 String bibLevel = mapBibLevel(componentLevel);
@@ -207,11 +207,11 @@ public class ComponentBuilder {
             if (dcmWorkPID != null)
                 componentWork.setDcmWorkPid(dcmWorkPID.toString());
 
-            mapContainer(componentWork, fieldsMap);
+            mapContainer(componentWork, fieldsMap, uuid);
         } catch (EADValidationException e) {
             throw e;
         } catch (Exception e) {
-            throw new EADValidationException("Failed to extract metadata for component work %s (Archives Space ID: %s)", e, componentWork.getObjId(), (uuid == null)?"":uuid);
+            throw new EADValidationException("FAILED_EXTRACT_GENERIC", e, componentWork.getObjId(), (uuid == null) ? "" : uuid);
         }
     }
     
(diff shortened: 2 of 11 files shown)


src/amberdb/model/builder/CollectionBuilder.java
```diff
@@ -90,7 +91,7 @@ public static void createCollection(Work collectionWork, JsonNode collectionCfg,
         }
         
         // initializing the parser
-        try (InputStream fileStream = eadFile.openStream()) {
+        try (InputStream fileStream = new BOMInputStream(eadFile.openStream(), false)) {
```
MingtaoSun: @danielbaker , so what is the purpose of using BOMInputStream? I have never used it before.

src/amberdb/model/builder/EADDuplicateValidator.java
```diff
@@ -0,0 +1,41 @@
+package amberdb.model.builder;
+
+import com.google.common.base.Joiner;
+
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Set;
+
+public class EADDuplicateValidator {
```
AaronCMuller: I don't think you're gaining anything making this a class with no effective properties, and creating an object that's immediately going to be thrown away.
I think it's fine for the validate method to be static.

src/amberdb/model/builder/EADDuplicateValidator.java
```diff
@@ -0,0 +1,41 @@
...
+
+    public EADDuplicateValidator(XmlDocumentParser parser) {
+        this.parser = parser;
+    }
+
+    private XmlDocumentParser parser;
+
+    public void validate() {
```
AaronCMuller: No throws declaration or javadoc? Without this, how are you supposed to know what this actually does without looking at the code.

src/amberdb/model/builder/ComponentSubUnitBuilder.java
```diff
@@ -0,0 +1,38 @@
+package amberdb.model.builder;
+
+import amberdb.enums.BibLevel;
+import amberdb.enums.SubUnitType;
+import amberdb.model.EADWork;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ComponentSubUnitBuilder {
```
AaronCMuller: A bibId is not a subUnit

AaronCMuller: bibLevel rather, not bibId

src/amberdb/model/builder/ComponentBuilder.java
```diff
@@ -237,28 +225,16 @@ protected static void extractDateRange(EADWork work, String uuid, String dateRan
...
         String containerParent = fieldsMap.get("container-parent");
         String containerNumber = fieldsMap.get("container-number");
         String containerLabel = fieldsMap.get("container-label");        
         String containerType = fieldsMap.get("container-type");
-        String componentWorkUUID = (componentWork.getLocalSystemNumber() == null)? "" : componentWork.getLocalSystemNumber();
 
         if (containerNumber == null || containerNumber.trim().isEmpty()) return;
         if (containerType == null || containerType.trim().isEmpty()) {
-            throw new EADValidationException("MISSING_CONTAINER_TYPE", componentWork.getObjId(), componentWorkUUID);
+            throw new EADValidationException("MISSING_CONTAINER_TYPE", componentWork.getObjId(), uuid);
```
terenceingram: Just a question does the string MISSING_CONTAINER_TYPE match the enum message in Banjo? Just wondering about this as I haven't really looked at this part of the code.

src/amberdb/model/builder/CollectionBuilder.java
terenceingram: I am also keen to learn about it. Is that part of the commons-io librar?

src/amberdb/model/builder/ComponentSubUnitBuilder.java
```diff
@@ -0,0 +1,38 @@
...
+    static final Logger log = LoggerFactory.getLogger(ComponentSubUnitBuilder.class);
+
+    public void setSubUnitAndBibLevelFields(EADWork componentWork, String uuid, String componentLevel) {
```
AaronCMuller: I'm a bit leery about methods with side-effects. Are you using Builder is the GoF sense? I would like to see this method return componentWork, or have a proper builder.

src/amberdb/model/builder/ComponentSubUnitBuilder.java
AaronCMuller: As with the UUIDValidator class, I think it is not useful to have a class with no real properties that is constructed and then discarded.

scoen:
i think that the banjo enum with:
`FAILED_EXTRACT_CONTAINERS("Failed to map containers for the work %s (Archives Space ID: %s)")`
now requires 2 strings passed rather than the one string it required previously. So i suspect until the amber change is merged the banjo change might bomb on encountering this exception type.

src/amberdb/model/builder/CollectionBuilder.java
AaronCMuller: Yes. Basically what it does is read the 'byte order mark' of a stream, and does stuff based on what it finds.

BOM's are essentially a leader for UTF files.

What this line is doing is constructing a stream that excludes the BOM from the stream.

AaronCMuller:
@terenceingram @scoen 

To be honest, I have no idea why this code is even in amber. It should really be in Banjo, then we can avoid this silliness of having error messages defined for something in another module.

src/amberdb/model/builder/ComponentSubUnitBuilder.java
```diff
@@ -0,0 +1,38 @@
...
+        if (componentLevel != null && !componentLevel.isEmpty()) {
+            log.debug("component work " + componentWork.getObjId() + ": componentLevel: "
```
AaronCMuller: I know this wasn't written by you, but could you change this to use interpolation rather than cat'ing

AaronCMuller:
Other than my concerns about the refactoring that has occurred, the general content of this PR is good.

src/amberdb/model/builder/CollectionBuilder.java
danielbaker: As Aaron said, it ignores some bytes at the beginning of UTF-8 files which windows occasionally adds. Java is unable to handle them, and one of the sample files Jenny provided had this issue

src/amberdb/model/builder/EADDuplicateValidator.java
danielbaker: While I agree that you gain nothing by instantiating, I don't see any advantages of making it static either other than saving a few chars. 

I generally prefer keeping things non-static because it makes things easier to unit test (though not in the state the tests for CollectionBuilder are in). Happy to change it to a static method if you think it has advantages, or if we have some code-style rules I am unaware of.

src/amberdb/model/builder/EADDuplicateValidator.java
danielbaker: I generally don't add javadoc to classes/methods unless they are being exposed as part of a public API, which this class is not. I have tried to name the classes/method in such a way that javadoc is unnecessary.

But I am happy to add javadoc if you feel it will add the the readability or if we have a code standard we are trying to stick to.

src/amberdb/model/builder/ComponentSubUnitBuilder.java
danielbaker: bibLevel appears to be calculated from the subunit, which is why I have put them together in this class. I have tried to indicate to the reader what is happening in the method name (setSubUnitAndBibLevelFields). 

Having said that I have no idea on the relative importance of the bibLevel vs SubUnit fields in this domain. If the BibLevel is more important I would probably rename this slightly

src/amberdb/model/builder/ComponentSubUnitBuilder.java
danielbaker: No this is not a builder in the GoF sense, I was just keeping the naming consistent with the Colleciton/Component Builders. But I agree that returning the componentWork would be more idiomatic, I will make this change

src/amberdb/model/builder/ComponentSubUnitBuilder.java
danielbaker: Will do

src/amberdb/model/builder/ComponentBuilder.java
danielbaker: Yes it does. Its messy

danielbaker:
I agree with Aaron that all the code to do with processing this file upload should sit in the same repo. Having it in separate repos probably added a few hours of dev time to this because it was difficult to follow the code path.

Do we have some way to track tech debt like this?

src/amberdb/model/builder/EADDuplicateValidator.java
AaronCMuller: We have no real coding standard.

I'm happy if you just add in a throws declaration.

src/amberdb/model/builder/EADDuplicateValidator.java
AaronCMuller: I'm happy for this to be left as is (I forgot about Mockito's issues with statics, etc). 

After thinking about this a lot, I think what bothers me about this is more to do with Java than anything you've done.

src/amberdb/model/builder/ComponentSubUnitBuilder.java
AaronCMuller: Maybe I'm blind, but that doesn't look like the case?

Let's have a chat about this tomorrow. I can also give you some information about subTypes and bibLevels, which is useful to know.

src/amberdb/model/builder/EADDuplicateValidator.java
MingtaoSun: @danielbaker, why non-static is easier to unit test?

src/amberdb/model/builder/CollectionBuilder.java
shuangzhou: would be nice and easier if we can check this during the file upload stage for an ead job, and save the validated file content if it's possible, should confirmation before proceed though.

src/amberdb/model/builder/EADDuplicateValidator.java
danielbaker: @MingtaoSun If you want to stub/mock something out in a unit test its more difficult to do if the method is static. The mocking library were using (Mockito) cannot do it. There are libraries you can use to do it (PowerMock) when its required.

MingtaoSun: Yep, it makes sense. :+1:

src/amberdb/model/builder/CollectionBuilder.java
AaronCMuller: I thinik it's fine to leave as is for now.

PR #387, opened 2015-08-21
Bs dscs 539 fa copy role

bsingh-nla:
10 new copy roles added finding aid 1, finding aid 2 ...finding aid 10

@scoen , @AaronCMuller , @ninhnguyen , @shuangzhou can you please look.


AaronCMuller:
I have rather strong feelings about this pull request. If possible I'd like to talk about this with Kavinga and Terence.

bsingh-nla:
@terenceingram , @AaronCMuller, @scoen Shaun had strong feelings too, it looks like we need to talk about it!

AaronCMuller:
@shuangzhou Why did you merge this?

terenceingram:
@bsingh-nla Can you revert this until we sort out the best approach?

shuangzhou:
@AaronCMuller I merged the change as the code looked ok, now I see that there's reason of concern for this, happy for it to be reverted.

bsingh-nla:
@terenceingram we are not able to revert it from github. I'll made code change.

PR #388, opened 2015-08-24
DSCS-221

yetti:
@MingtaoSun @bsingh-nla 

Renamed `hasCopyType()` to `hasCopyRole()`. I thought it would be less likely to get confused with the `getCopyType()` method on the `Copy` object.


pom.xml
```diff
@@ -3,7 +3,7 @@
 
   <groupId>au.gov.nla</groupId>
   <artifactId>amberdb</artifactId>
-  <version>1.2.6-SNAPSHOT</version>
+  <version>1.2.7-SNAPSHOT</version>
```
MingtaoSun: I thought we are not doing it any more for minor changes

yetti: I wasn't told that...

bsingh-nla: Based on the discussion I had with Shaun, if new change breaks existing code, we should update version number.

MingtaoSun: OK, sounds good

PR #390, opened 2015-08-24
Ms1207

MingtaoSun:
@shuangzhou @AaronCMuller @bsingh-nla 
- add a new field to Work and Copy respectively.
- add two new copy roles
- Change the order numbers so there is gaps for inserting new records easily. (The ord property has no business significance right?)


MingtaoSun:
@bsingh-nla @shuangzhou @scoen 
conflict resolved.

bsingh-nla:
looks good to me.

PR #393, opened 2015-08-25
ms1207: move html field from copy to work

MingtaoSun:


shuangzhou:
@MingtaoSun looks good to me.

PR #394, opened 2015-08-27
Sc paged child results

scoen:


src/amberdb/query/WorkChildrenQuery.java
```diff
@@ -183,4 +187,36 @@ public Integer getTotalChildCount(Long workId) {
...
+                    .bind("workVal", AmberProperty.encode("Work"))
+                    .bind("pageVal", AmberProperty.encode("Page"))
+                    .bind("eadWorkVal", AmberProperty.encode("EADWork"))
+                    .map(ByteArrayMapper.FIRST).list();
+        }
+
+        List<CopyRole> copyRoles = new ArrayList<CopyRole>();
+        for (byte[] bytes : copyRoleCodes) {
+            String code = (String) AmberProperty.decode(bytes, DataType.STR);
+            copyRoles.add(CopyRole.fromString(code));
```
MingtaoSun: So we assume all the code saved in db can be converted to CopyRole enum

scoen: i think we should be able to assume that. I think the method should fail on that sort of bad data rather than letting it persist for now.

PR #403, opened 2015-09-10
update json serialization.

shuangzhou:
update json serialization.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 AmberDB                       
 =======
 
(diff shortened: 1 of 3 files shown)


terenceingram:
@scoen Are you happy with this change.

scoen:
it would have been better to do it on the SerialiseToJSON method so it works for all methods that call it. I'll look at doing that.

scoen:
Actually, Shuang's fix is good - creating a HashSet won't take null as an arguement

PR #404, opened 2015-09-10
Bs dscs 106 ack enums

bsingh-nla:
enums added for work acknowledgement

@scoen can you please look.


scoen:
looks good. I'm curious why one test run failed, but i expect it's travis funnies.

PR #405, opened 2015-09-10
add empty list guard to json serialisation

scoen:
@shuangzhou 
After seeing the enhancement you put in earlier to ensure uniqueness for Subjects during JSON serializing, i've also added a guard to the serialiseToJSON method to prevent empty lists being saved. Would you take a look ?


shuangzhou:
Good idea.

PR #406, opened 2015-09-10
Sc large txns

scoen:
Committing very large graphs to amber will now batch the marshaling of graph elements (vertices and edges and their properties) to prevent the object creation involved with marshaling from using too much memory. 

It also slightly improves the speed of committing in many cases through the creation of indexes on the session tables which get used by queries during commit, and modification to the queries themselves to ensure the indexes get used.

The big commit threshold has been set to 50000 vertices+edges. This is somewhat arbitrary and just based on past experience of where commits have failed due to running out of memory. I don't think it's necessary to make it configurable (yet). 

You can call the commitBig method explicitly (ie: it's public) but the existing commit methods will automatically use it if the threshold conditions are met.  

I updated the snapshot version due to the significance of the change - rolling back from it could cause processes that rely on this feature to fail.


scoen:
I haven't committed tests for this feature, as they do take a very long time to actually test the cases where it is useful. It has been tested for 190000 elements (which have 1000000+ properties altogether) in an otherwise empty jvm. I recognise that this is a pretty artificial case, and normally alot more would be going on.

JonathanShaw:
This looks really handy.  Looking over the changes now

shuangzhou:
Awesome from the first impression, just wondering whether it's a good idea to have checkpoints also for such large transactions to swiftly recover from runtime disruption.

scoen:
Cheers for the feedback Shuang. The checkpoint idea is good too, but for sake of simplicity I won't do that yet - the complexity of figuring out where to recover from is too much for me at the moment :-) It already does do some backoff retries if the writing to the actual amberdb tables fails, and if the writing to session tables fails it's simpler just to redo them as it won't pollute the real graph tables (although admittedly it could leave a bunch of useless entries in the session tables)

PR #409, opened 2015-09-15
Bs dscs 539 fa update

bsingh-nla:
@scoen I'm back with FINDING_AID copy role 1 to 10 :)


MingtaoSun:
Looks like I should have made the gap 100 instead of 10 :smile:

bsingh-nla:
@MingtaoSun  :wink: Probably that's not enough...if i add finding aid copy role 1 to 100

AaronCMuller:
Why?

bsingh-nla:
@AaronCMuller you mean why finding aid 1 to 10 again?

AaronCMuller:
Yes.

bsingh-nla:
Probably Kavinga can answer it better.

JonathanShaw:
I'm probably just being slow here but what was the reason we aren't dealing with multiple Finding Aid copies? I thought AmberDb had been updated to allow this

bsingh-nla:
Again, not sure. Kavinga is the right person to answer this.

PR #412, opened 2015-09-16
Ms229: Modify CameraData fields

MingtaoSun:
@bsingh-nla 
Some of the cameraData fields cannot display correctly.

### DIFF
--- a/src/amberdb/model/CameraData.java
+++ b/src/amberdb/model/CameraData.java
@@ -6,40 +6,40 @@ import com.tinkerpop.frames.modules.typedgraph.TypeValue;
 @TypeValue("CameraData")
 public interface CameraData extends Description {
     @Property("exposureTime")
-    public String getexposureTime();
+    public String getExposureTime();
     
     @Property("exposureTime")
-    public void setexposureTime(String exposureTime);
+    public void setExposureTime(String exposureTime);
     
     @Property("exposureFNumber")
-    public String getexposureFNumber();
+    public String getExposureFNumber();
     
     @Property("exposureFNumber")
-    public void setexposureFNumber(String exposureFNumber);
+    public void setExposureFNumber(String exposureFNumber);
     
     @Property("exposureMode")
-    public String getexposureMode();
+    public String getExposureMode();
     
     @Property("exposureMode")
-    public void setexposureMode(String exposureMode);
+    public void setExposureMode(String exposureMode);
     
     @Property("exposureProgram")
-    public String getexposureProgram();
+    public String getExposureProgram();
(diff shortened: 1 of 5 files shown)


src/amberdb/model/CameraData.java
```diff
@@ -6,40 +6,40 @@
 @TypeValue("CameraData")
 public interface CameraData extends Description {
     @Property("exposureTime")
-    public String getexposureTime();
+    public String getExposureTime();
```
bsingh-nla: you might have to modify banjo as well...where these methods are used.

src/amberdb/model/Description.java
```diff
@@ -34,5 +37,18 @@
         public Set<String> getPropertyKeySet() {
             return this.asVertex().getPropertyKeys();
         }
+
+        @Override
+        public long getId() {
+            return toLong(asVertex().getId());
+        }
+
+        public long toLong(Object x) {
```
scoen: maybe make this private. Shouldn't be used as a utility function outside of the Description

scoen:
You'll Need to modify amberdb.models.CameraData too to call the correct methods

MingtaoSun:
@AaronCMuller , Aaron, what do you mean by 'Please don't push to amberdb without a pull request.'
Isn't this one a pull request?

src/amberdb/model/CameraData.java
MingtaoSun: Yes, I have modified them. haven't put in a pull request yet. That's why I up the pom version, so that it won't cause banjo to break.

AaronCMuller:
Oh sorry, ignore that comment then. The way it looked from the github stream looked like you directly pushed to master without a PR.

src/amberdb/model/CameraData.java
MingtaoSun: @bsingh-nla , yeah, better not to merge until I put in the banjo PR.

src/amberdb/model/Description.java
```diff
@@ -34,5 +37,18 @@
         public Set<String> getPropertyKeySet() {
             return this.asVertex().getPropertyKeys();
         }
+
+        @Override
+        public long getId() {
```
scoen: Can i just ask why you want the id ?

MingtaoSun: To get the history info
VersionedVertex versionedVertex = amberSession.getAmberHistory().getVersionedGraph().getVertex(cameraData.getId());

scoen: Fair enough

PR #413, opened 2015-09-17
Added utility method that checks if a work is a valid sound work

yetti:
@danielbaker @AaronCMuller I've added a method on the Work object to check if the collection is "nla.oh".

### DIFF
--- a/src/amberdb/model/Work.java
+++ b/src/amberdb/model/Work.java
@@ -1212,6 +1212,9 @@ public interface Work extends Node {
     @JavaHandler
     public boolean hasCopyRole(CopyRole role);
 
+    @JavaHandler
+    public boolean isValidSoundWork();
+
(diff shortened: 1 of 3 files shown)


brycollins:
I think its a bit misleading to assume its a valid sound file just because its in the nla.oh collection.
Shouldnt this be more like isInCollection("nla.oh") ?? Then it can apply to all works and collections.

src/amberdb/model/Work.java
```diff
@@ -1732,5 +1735,10 @@ public boolean hasMasterCopy() {
         public boolean hasCopyRole(CopyRole role) {
             return getCopy(role) != null;
         }
+
+        @Override
+        public boolean isValidSoundWork() {
+            return "nla.oh".equalsIgnoreCase(getCollection());
```
danielbaker: I was thinking you would add the bit in here about the number of aliases being exactly 1 and no other works having the same alias

PR #414, opened 2015-09-17
Check if work is a valid sound work

yetti:
OK, let's try this again. @danielbaker @AaronCMuller

### DIFF
--- a/src/amberdb/model/Work.java
+++ b/src/amberdb/model/Work.java
@@ -12,6 +12,7 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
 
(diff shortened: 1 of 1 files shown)


danielbaker:
I like it!

JonathanShaw:
I like this,  you can probably refactor a couple of things to use this if you're feeling like it. Off the top of my head spreadsheet upload, sound consignment utils and the work edit page all would be able to use this.

yetti:
Yep, looking at sound consignment and spreadsheet upload atm since I'm already working on sound consignment.

JonathanShaw:
thanks :)

PR #415, opened 2015-09-18
Bs dscs 106 ack sorting

bsingh-nla:
@scoen , @AaronCMuller , @yetti , @danielbaker can you please look.

### DIFF
--- a/src/amberdb/model/Work.java
+++ b/src/amberdb/model/Work.java
@@ -445,7 +445,7 @@ public interface Work extends Node {
             final Double weighting, final Date dateOfAck, final String urlToOriginial);
 
     @JavaHandler
-    public List<Acknowledge> getAcknowledgementsByWeighting();
+    public List<Acknowledge> getOrderedAcknowledgements();
     /**
      * This property is encoded as a JSON Array - You probably want to use
      * getSeries to get this property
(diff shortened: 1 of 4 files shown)


src/amberdb/model/Work.java
```diff
@@ -1230,6 +1234,19 @@ public Acknowledge addAcknowledgement(final Party party, final String ackType, f
...
+            List<Acknowledge> list = Lists.newArrayList(getAcknowledgements());
+
+            Collections.sort(list, new Comparator<Acknowledge>() {
+                public int compare(final Acknowledge object1, final Acknowledge object2) {
+                    return object1.getWeighting().compareTo(object2.getWeighting());
+                }
+            });
+
+            return list;
+        }
```
yetti: Seems like something that ideally should probably be done by the DB with an orderby query, but not sure if this is easily done in a graph.

yetti:
Looks good apart from my other comment.

src/amberdb/model/Work.java
scoen: Agreed - it should be done by the graph query language. Unfortunately, although the Gremlin query language used by tinkerpop implementations allows for returning Edges sorted on attributes, the GremlinGroovy java annotations available to tinkerpop can only return Vertices. So while the Edges can be sorted before being returned from a GremlinGroovy query, but they aren't the correct class and you have to loop through the results to cast them to the desired class which kinda defeats the purpose. So, easiest solution is (unfortunately) just to sort the returned Edges [sigh].

PR #416, opened 2015-09-22
query to retrieve a Work's Sections (articles and chapters)

scoen:
Need this query to make Banjo show Work page more efficient - currently it queries all children twice to get the articles and chapters for display. This takes a long time for Works with many children. I probably need to refactor getting all of a work at some point, but in the meantime this should do the trick.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 AmberDB                       
 =======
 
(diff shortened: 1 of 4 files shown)


scoen:
anyone interested ?

MingtaoSun:
:+1:

bsingh-nla:
looks good

PR #417, opened 2015-09-22
Add additional copy role, and new method for retrieving copies.

AaronCMuller:
@scoen @JonathanShaw


JonathanShaw:
might want to update the validateCopyRoleEnum test.

Otherwise loks good to me

scoen:
me too

scoen:
I suspect the test will still fail, because it counts the number of copyroles at the end :-(

PR #419, opened 2015-09-22
Sc fix tests

scoen:


scoen:
Don't merge yet - i want to see how tests go first

scoen:
Hmmmmm, I may well have just got lucky.
However if other tests fail intermittently i might try converting them to use the abstract test class and see if that makes things better.

PR #420, opened 2015-09-29
Sc works query

scoen:
@danielbaker Sorry it took so long, but here's a query for works that takes a list of work Ids.

call like `List<Work> works = WorksQuery.getWorks(amberSession, ids);`


danielbaker:
Thanks Shaun looks good :)

PR #422, opened 2015-09-30
Sc update multipart query

scoen:
Also contains a lot of warning fixes raised by intelliJ like interface method signatures not requiring 'public'


shuangzhou:
Looks good.

PR #423, opened 2015-09-30
Add ability to retrive copy index.

AaronCMuller:
@scoen @JonathanShaw

### DIFF
--- a/src/amberdb/model/Copy.java
+++ b/src/amberdb/model/Copy.java
@@ -16,6 +16,7 @@ import java.util.List;
 import java.util.Map;
 
 import amberdb.relation.*;
+import com.google.common.base.Predicate;
 import com.google.common.collect.Iterables;
 import org.apache.commons.lang.StringUtils;
 import org.apache.tika.Tika;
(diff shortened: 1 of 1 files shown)


src/amberdb/model/Copy.java
```diff
@@ -779,5 +783,18 @@ else if (fmt.equals("audio"))
...
         }
+
+        @Override
+        public int getCurrentIndex() {
+            final Copy copy = this;
+            int ind = Iterables.indexOf(getWork().getOrderedCopies(CopyRole.fromString(getCopyRole())),
+                                        new Predicate<Copy>() {
+                                            @Override
+                                            public boolean apply(Copy iCopy) {
+                                                return copy.getId() == iCopy.getId();
```
scoen: .equals ?

AaronCMuller: <code>abstract public long getId();</code>
I don't think so?

scoen: Fair enough. i hadn't looked at method signature - thought was Long returned.

PR #424, opened 2015-10-02
New query to get copies by id list and load their works

scoen:
The returned list is of the Copies, but their Works have also been loaded to the local graph for quicker access.
@danielbaker 
 Is this useful ?

### DIFF
--- a/src/amberdb/query/WorksQuery.java
+++ b/src/amberdb/query/WorksQuery.java
@@ -1,9 +1,14 @@
 package amberdb.query;
 
 import amberdb.AmberSession;
+import amberdb.graph.AmberQuery;
+import amberdb.graph.BranchType;
+import amberdb.model.Copy;
 import amberdb.model.Work;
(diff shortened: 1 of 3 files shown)


src/amberdb/query/WorksQuery.java
```diff
@@ -1,9 +1,14 @@
...
+import amberdb.graph.BranchType;
+import amberdb.model.Copy;
 import amberdb.model.Work;
 import java.util.ArrayList;
 import java.util.List;
+
+import com.tinkerpop.blueprints.Direction;
 import com.tinkerpop.blueprints.Vertex;
 
 public class WorksQuery {
```
AaronCMuller: I like the idea of a WorksQuery thingy, so we can not overload Work so much :+1:

AaronCMuller:
I'm happy with this.

@danielbaker Can you confirm that you still want this functionality?

PR #426, opened 2015-10-08
ms1105: add SiblingEdtions

MingtaoSun:
@terenceingram 
added the new class and field to handle the 'sibling' editions (one map grid has multiple PIs)
I think the info is better extracted and saved into 'interactive index' copy file, rather than re-parse the html at dl-repo service.
So now the json file looks like this
{
"html":"< map name=\"mapname\"> < area >  .. </area > < /map >",
"siblingEditionsList":[{"objectIds":["nla.obj-19192240","nla.obj-19192350"]}],
"areas":[{"objectId":"nla.obj-19195544","neighbours":[{"direction":"E","objectId":"nla.obj-19194664"},
}

### DIFF
--- a/src/amber/interactiveindex/InteractiveIndexCopy.java
+++ b/src/amber/interactiveindex/InteractiveIndexCopy.java
@@ -1,5 +1,6 @@
 package amber.interactiveindex;
 
+import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.lang.StringUtils;
 
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


terenceingram:
Interesting solution. We will still need to get each siblings title, but that can be loaded on the Tarkine Client as Ajax.

terenceingram:
@scoen Are you happy with these Amber changes?

MingtaoSun:
@terenceingram , title and other fields will still come from dl-repo. Stay tuned....

scoen:
Yeah i think it looks ok

PR #428, opened 2015-10-09
ms1138: get all sibling edtion object ids including itself

MingtaoSun:
@terenceingram @scoen 
- need to display all the sibling editions (including itself) in the dropdown


MingtaoSun:
@bsingh-nla @AaronCMuller @JonathanShaw @shuangzhou @ninhnguyen , happy to merge? need to get the dl-repo-service to pick up the change so that i can test in Tarkine.

PR #430, opened 2015-10-12
ms1106: add 3 new fields

MingtaoSun:
@scoen @shuangzhou @bsingh-nla 
http://ourweb.nla.gov.au/apps/jira/browse/DSCS-1106


bsingh-nla:
code looks good...just waiting for CI to finish check

PR #432, opened 2015-10-14
Sc wrap retry and delivered on

scoen:
2 fixes in this pull request (see first 2 commits)
first commit fixes bad wrapping of commit retries that was causing duplicate vertices in amberdb
second adds the 'deliveredOn' relationship to the loading of works into memory
@yetti 
Yetrina, would you take a look at the second one at least ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 AmberDB                       
 =======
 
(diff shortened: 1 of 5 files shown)


yetti:
That looks alright to me :+1:

PR #434, opened 2015-10-14
Fix relation when getting a delivery work parent edge

yetti:
@scoen can you please have a look?

### DIFF
--- a/src/amberdb/model/Work.java
+++ b/src/amberdb/model/Work.java
@@ -1000,10 +1000,10 @@ public interface Work extends Node {
     void removeDeliveryWorkParent(final Work interview);
 
     @Incidence(label = DeliveredOn.label, direction = Direction.IN)
-    Iterable<ExistsOn> getDeliveryWorkParentEdges();
+    Iterable<DeliveredOn> getDeliveryWorkParentEdges();
 
     @JavaHandler
-    ExistsOn getDeliveryWorkParentEdge();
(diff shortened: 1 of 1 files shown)


src/amberdb/model/Work.java
```diff
@@ -1000,10 +1000,10 @@ Acknowledge addAcknowledgement(final Party party, final String ackType, final St
     void removeDeliveryWorkParent(final Work interview);
 
     @Incidence(label = DeliveredOn.label, direction = Direction.IN)
-    Iterable<ExistsOn> getDeliveryWorkParentEdges();
+    Iterable<DeliveredOn> getDeliveryWorkParentEdges();
```
MingtaoSun: @yetti , @scoen , not sure if it has implications on history tab, namely this line
Iterable<VersionedEdge> versionedEdges = versionedVertex.getEdges(Direction.IN, "deliveredOn");
Will it still return the same result?

yetti: I think the version is still going to function normally, but the label it uses will be different. @scoen please correct me if I'm wrong.

scoen: This will depend on whether we have legacy data using existsOn edges to model the deliveredOn relationship data. If so, the data needs correcting. 
I can't remember how the history tab is implemented, so i'm unsure what the consequences might be. I'd need to take a look.

scoen:
I think the versioning won't be affected adversely.

PR #435, opened 2015-10-15
Sc fix jp2 derive

scoen:
@ninhnguyen would you take a look ? I think this does the trick.


ninhnguyen:
@scoen  Looks good to me!
Shuang and I don't know why it failed the test shouldRemoveMultipleAcknowledgementsFromAWork?

scoen:
@ninhnguyen @shuangzhou 
Good question! I fear it's related to the way tests are set up and torn down. Before merging i might try making that test less frail. I'll get back to you.

scoen:
@ninhnguyen @shuangzhou 
It's passed the tests now. Whether it was my change or just because of the weather(:-) i don't know.

PR #436, opened 2015-10-19
ms1342: bulk edit bib level

MingtaoSun:
http://ourweb.nla.gov.au/apps/jira/browse/DSCS-1342
@AaronCMuller @shuangzhou @bsingh-nla @scoen 
- Need the order to determine the lowest bib level of multiple works for bulk editing

### DIFF
--- a/src/amberdb/enums/BibLevel.java
+++ b/src/amberdb/enums/BibLevel.java
@@ -12,11 +12,11 @@ public enum BibLevel{
     NO_BIB_LEVEL("", 0);
 
     private String code;
-    private Integer order;
+    private Integer level;
 
-    private BibLevel(String code, Integer order) {
+    private BibLevel(String code, Integer level) {
         this.code = code;
-        this.order = order;
+        this.level = level;
     }
 
     public static BibLevel fromString(String code) {
@@ -34,8 +34,8 @@ public enum BibLevel{
         return code;
     }
     
-    public Integer order() {
-        return order;
+    public Integer level() {
+        return level;
     }
     
     public static List<String> list() {
--- a/src/amberdb/enums/BibLevel.java
+++ b/src/amberdb/enums/BibLevel.java
@@ -5,11 +5,10 @@ import java.util.List;
 
 public enum BibLevel{
     
-    COMPONENT("Component", 5), 
-    SET("Set", 3), 
-    ITEM("Item", 2), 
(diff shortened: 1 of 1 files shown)


AaronCMuller:
I'm going to review this alongside the banjo PR for this, so please don't merge this until that time.

src/amberdb/enums/BibLevel.java
```diff
@@ -3,18 +3,20 @@
...
-    PART("Part"),
-    SECTION("Section");
+    SET("Set", 4), 
+    ITEM("Item", 3), 
+    PART("Part", 2),
+    SECTION("Section", 1),
+    NO_BIB_LEVEL("", 0);
 
     private String code;
+    private Integer order;
```
AaronCMuller: I think this should be level, not order since it refers to it's level in the hierarchy, rather than an ordering.

src/amberdb/enums/BibLevel.java
```diff
@@ -3,18 +3,20 @@
...
+public enum BibLevel{
     
-    COMPONENT("Component"), 
-    SET("Set"), 
-    ITEM("Item"), 
-    PART("Part"),
-    SECTION("Section");
+    SET("Set", 4), 
+    ITEM("Item", 3), 
+    PART("Part", 2),
```
AaronCMuller: I think both part and section should have the same number (i.e. 1)

AaronCMuller:
Happy for this to be merged, once you've addressed the comments.

src/amberdb/enums/BibLevel.java
MingtaoSun: I think section should be lower than part. If the lowest level is part, you can still set the bib level to part. However, if the lowest level is section, you cannot set any bib level.

AaronCMuller: Yep, I see what you mean.

src/amberdb/enums/BibLevel.java
```diff
@@ -3,18 +3,20 @@
...
+    SECTION("Section", 1),
+    NO_BIB_LEVEL("", 0);
 
     private String code;
+    private Integer level;
 
-    private BibLevel(String code) {
+    private BibLevel(String code, Integer level) {
```
MingtaoSun: @AaronCMuller Renamed

PR #437, opened 2015-10-19
Change to implement the digital book automatic ingest - ddg-105

sjacob:
This change is to add a new SubType called "Cover". The subtype is used to describe the cover art file received as part of a digital deposit. @JonathanShaw @terenceingram @scoen @AaronCMuller


AaronCMuller:
You might want to talk to Jon or Terence about this.

SubType for the most part should no longer be used.

scoen:
@terenceingram 
Are we still relying on subtype ? I guess so, but just want to check before i merge.

terenceingram:
@scoen @AaronCMuller @sjacob We don't rely on subtype. It should be totally deprecated. The only use currently is for Sections (articles and chapters). Otherwise we look at Collection, Bib Level and Form to make decisions. So this needs to be changed.

scoen:
Not adding more subtypes. Hopefully all dependencies on subtypes will be removed in the near future.

sjacob:
Where is it documented that subtype is no longer used?

terenceingram:
@sjacob  It's not. It should have a @deprecated annotation above the method. We stopped using it well over a year ago and haven't had the time to actively remove it. Sorry I know this is frustrating.

:cc @scoen

PR #440, opened 2015-10-21
Ms1134: sort on field children reorder

MingtaoSun:
http://ourweb.nla.gov.au/apps/jira/browse/DSCS-1134
@scoen @AaronCMuller @danielbaker 
- Add a new query based on the getChildRange() to return a subset of children sorted by certain field
- @scoen , the original query wasn't changed. Just refactored a bit to fit in my query
- Unfortunately, the new query isn't unit testable due to
1) ISNULL() function is not recognised by jdbi
2) The order by field must be in the SELECT clause in jdbi (while not required by MySQL)
So the query has been tested in Banjo instead
- The banjo PR is coming out soon. Please review along with the banjo PR.

### DIFF
--- a/src/amberdb/query/WorkChildrenQuery.java
+++ b/src/amberdb/query/WorkChildrenQuery.java
@@ -1,5 +1,20 @@
 package amberdb.query;
 
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.apache.commons.codec.binary.Hex;
+import org.apache.commons.lang.StringUtils;
+import org.skife.jdbi.v2.Handle;
+import org.skife.jdbi.v2.util.ByteArrayMapper;
+import org.skife.jdbi.v2.util.IntegerMapper;
+
+import com.tinkerpop.blueprints.Vertex;
+
 import amberdb.AmberSession;
 import amberdb.enums.CopyRole;
 import amberdb.graph.AmberProperty;
@@ -7,20 +22,7 @@ import amberdb.graph.AmberQueryBase;
 import amberdb.graph.DataType;
 import amberdb.model.Section;
 import amberdb.model.Work;
-
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-
-import org.skife.jdbi.v2.Handle;
-import org.skife.jdbi.v2.util.ByteArrayMapper;
-import org.skife.jdbi.v2.util.IntegerMapper;
-
-import com.tinkerpop.blueprints.Vertex;
-
-import org.apache.commons.codec.binary.Hex;
-import org.apache.commons.lang.StringUtils;
+import amberdb.sort.SortItem;
 
 public class WorkChildrenQuery extends AmberQueryBase {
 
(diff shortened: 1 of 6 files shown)


src/amberdb/query/WorkChildrenQuery.java
```diff
@@ -59,8 +60,20 @@ public WorkChildrenQuery(AmberSession sess) {
...
+        return getChildren(getAddChildrenWokSql(workId, start, num));
+    }
     
-    public List<Work> getChildRange(Long workId, int start, int num) {
+    public List<Work> getChildRangeSortBy(Long workId, int start, int num, final SortItem sortItem){
+        if (sortItem == null){
+            return getChildRange(workId, start, num);
+        }
+        List<Work> works = getChildren(getAddChildrenWorkSortBySql(workId, start, num, sortItem.fieldName(), sortItem.desc()));
+        Collections.sort(works, sortItem.compartor());
```
MingtaoSun: Need to sort the list again, because the order in the temp table v1 cannot be preserved.

MingtaoSun: @scoen , Hi Shaun, is it possible to sort the v1 table again at the end so that it doesn't need sorted by Java?

src/amberdb/query/WorkChildrenQuery.java
```diff
@@ -164,6 +165,39 @@ public WorkChildrenQuery(AmberSession sess) {
         }
         return children;
     }
+
+    private String getAddChildrenWokSql(Long workId, int start, int num) {
```
scoen: Wok ? :-)

AaronCMuller: Adding children to Woks is generally not condoned in this establishment.

src/amberdb/query/WorkChildrenQuery.java
scoen: spelling (compartor)

scoen:
Hi Ming,

There's a lot of sorting code in here. Just a thought - could you create the v1 table pulling in an extra column (the value from the property you want to sort on) then sort on that ? Would that eliminate the need for all the sort classes ?

MingtaoSun:
@scoen spelling mistakes have been fixed.

test/amberdb/sort/SortFieldTest.java
```diff
@@ -0,0 +1,49 @@
...
+        work1 = mock(Work.class);
+        work2 = mock(Work.class);
+        work3 = mock(Work.class);
+        work4 = mock(Work.class);
+        work5 = mock(Work.class);
+    }
+    
+    @Test
+    public void aliasSort() throws Exception{
+        when(work1.getAlias()).thenReturn(new ArrayList<String>());
```
danielbaker: When writing tests like this I would prefer to put all the `when` statements in the `@Before` and then write multiple tests which just have a single assertion. In its current state, if the test fails on a particular line for whatever reason (eg. line 39) its not immediately clear why its failing. If its in its own test with its own name its easier to determine why a test has failed

AaronCMuller:
Ming, I merged this so it doesn't break devel due to needing this. Please still address the feedback given.

scoen:
Eeek! i was working on a change. I should have said something

scoen:
can i revert ?

PR #441, opened 2015-10-21
Added Picture and Digital Publication form enum values.

sjacob:
Added Picture and Digital Publication form enum values that are used by works deposited under the new copyright legislation.


scoen:
:+1: looks ok to me

PR #445, opened 2015-10-22
ms1134: add getAddChildrenWorksSortBySql

MingtaoSun:
@scoen , Shaun, as discussed, I need to put a separate SQL in to sort the whole children first.


scoen:
That looks right

PR #447, opened 2015-10-23
Bs dscs 1240 copyrole

bsingh-nla:
@scoen , @MingtaoSun , @yetti , @AaronCMuller , @shuangzhou can you please look.

http://ourweb.nla.gov.au/apps/jira/browse/DSCS-1240

fix inconsistent copy role cases.


scoen:
:+1:

MingtaoSun:
I suggest not increasing the pom version for such minor change in the future.

bsingh-nla:
@MingtaoSun I was worried about breaking other apps with this change.

PR #450, opened 2015-10-26
Added a new fields to Work and File

sjacob:
This change adds a new "checksum generation date" field to File.  This is the date that the checksum was generated by either Banjo or an external application. The E-Deposit application will include this data when it submits files to Banjo

A new "vendor id" field has been added to Work. Vendor id is generated by the E-Deposit application and needs to be included in the data sent to Voyager.
@scoen

### DIFF
--- a/src/amberdb/model/Work.java
+++ b/src/amberdb/model/Work.java
@@ -430,6 +430,18 @@ public interface Work extends Node {
     
     @Property("sheetCreationDate")
     void setSheetCreationDate(Date sheetCreationDate);
+    
+    /**
+     * Get the vendor identifier that was assigned by the E-Deposit app.
+     */
+    @Property("vendorId")
+    String getVendorId();
+    
+    /**
+     * Set the vendor identifier that was assigned by the E-Deposit app.
+     */
+    @Property("vendorId")
+    void setVendorId(String id);
 
     @Adjacency(label = DescriptionOf.label, direction = Direction.IN)
     GeoCoding addGeoCoding();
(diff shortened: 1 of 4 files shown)


src/amberdb/model/Work.java
```diff
@@ -430,6 +430,18 @@
     
     @Property("sheetCreationDate")
     void setSheetCreationDate(Date sheetCreationDate);
+    
+    /**
+     * Get the vendor identifier that was assigned by the E-Deposit app.
+     */
+    @Property("vendorId")
```
terenceingram: @scoen I wonder if this should rather be stored in the ALIAS PI field? Just a thought. The only drawback is that it could be deleted :(

scoen:
looks fine to me

src/amberdb/model/Work.java
scoen: @sjacob @terenceingram 
Personally, I don't think we should treat vendor id as an alias pi - i feel like that would be overloading the meaning of alias pi and may have later modelling implications. Just my 2 cents.

sjacob: @terenceingram  
I agree with @scoen . VendorId is an identifier for the publisher who submitted the work, not an alternate identifier for this particular work.

terenceingram: @scoen @sjacob All good. Just checking :)

terenceingram: Is this Vendor ID going to be added to the index and searchable?

sjacob: According to http://ourweb.nla.gov.au/apps/jira/browse/DDG-108, yes it will be searchable and also will be passed to Voyager.

PR #455, opened 2015-11-03
[DSCS-1601] Adding Pre commit hooks and total duration to works

danielbaker:
Adding pre-commit hooks to Amber. These are setup in Banjo and are run every time a commit is made in amber

See this banjo PR for how this can be used: https://github.com/nla/banjo/pull/1697

### DIFF
--- a/src/amberdb/util/DurationUtils.java
+++ b/src/amberdb/util/DurationUtils.java
@@ -7,6 +7,9 @@ import org.joda.time.format.PeriodFormatterBuilder;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
 public class DurationUtils {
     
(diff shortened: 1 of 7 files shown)


src/amberdb/util/DurationUtils.java
```diff
@@ -37,4 +40,49 @@ public static String convertDuration(final String periodHHMMSSmm){
         }
         return newDuration;
     }
+
+    public static Float convertDurationToSeconds(String duration) {
```
danielbaker: Moved line-for-line from SoundFile.java

danielbaker:
@scoen @AaronCMuller @JonathanShaw @yetti

src/amberdb/util/DurationUtils.java
yetti: :+1:

yetti:
Looks ok to me

PR #456, opened 2015-11-04
ms1606: add width and height

MingtaoSun:
@terenceingram @kenjiwalter @scoen @bsingh-nla 
need width and height to display the image in browser widget in Tarkine

### DIFF
--- a/src/amber/interactiveindex/InteractiveIndexCopy.java
+++ b/src/amber/interactiveindex/InteractiveIndexCopy.java
@@ -8,11 +8,15 @@ import java.util.List;
 
 public class InteractiveIndexCopy {
     private String html;
+    private int width;
+    private int height;
     private List<SiblingEditions> siblingEditionsList = new ArrayList<>();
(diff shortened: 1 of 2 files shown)


scoen:
looks ok to me

src/amber/interactiveindex/InteractiveIndexCopy.java
```diff
@@ -8,11 +8,15 @@
 
 public class InteractiveIndexCopy {
```
AaronCMuller: Why doesn't InteractiveIndexCopy implement Copy?

MingtaoSun: Because it has nothing to do with the copy class. nor does it with 99% of any classes in amberdb. It shouldn't belong here in the first place. @scoen and I had a discussion and decided to temporarily put it here.

 It just holds information of the generated HTML, neighbor areas, and sibling editions, etc.. Perhaps it should be renamed. Anyway, I think @scoen is going to move this class to a separate project later. When he does, we can think of a better name.

PR #457, opened 2015-11-05
Remove IOExceptions thrown from Json deserialization.

AaronCMuller:
The rationale here is that if any exception occurs deserializing a json property, it probably means that the data somehow isn't actually json, and we should fail hard, rather than having to deal with IOExceptions everywhere.

This also removes outdated versions of jackson that were being used rather than the newer (already imported) versions.

@scoen

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,7 @@
 
   <groupId>au.gov.nla</groupId>
   <artifactId>amberdb</artifactId>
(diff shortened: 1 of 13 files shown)


MingtaoSun:
good idea :+1:

scoen:
agreed - good idea. Unfortunately currently causes tests to bomb.

PR #460, opened 2015-11-09
adding new subunit types

snezanam:
Adding the new subunit types in the lookups, @scoen, @shuangzhou  please take a look.


shuangzhou:
Looks good.

PR #466, opened 2015-11-11
ms1689: add new 'Event note' field

MingtaoSun:
@AaronCMuller , @scoen @shuangzhou 
http://ourweb.nla.gov.au/apps/jira/browse/DSCS-1689
This is the Voyager 518 field, similar to Finding Aid Note(555) and Restriction on Access (506)


bsingh-nla:
:+1:

PR #469, opened 2015-11-12
Updated Work.setOrder() so that it doesn't throw a NullPointerException

sjacob:
Updated Work.setOrder() so that it doesn't throw a NullPointerException when the work doesn't have a parent.

@scoen @AaronCMuller


AaronCMuller:
That looks good.

PR #471, opened 2015-11-16
DanB: Strip out milliseconds from Duration utils

danielbaker:
@scoen

### DIFF
--- a/src/amberdb/util/DurationUtils.java
+++ b/src/amberdb/util/DurationUtils.java
@@ -42,20 +42,17 @@ public class DurationUtils {
     }
 
     public static Float convertDurationToSeconds(String duration) {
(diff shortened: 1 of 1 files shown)


src/amberdb/util/DurationUtils.java
```diff
@@ -42,20 +42,17 @@ public static String convertDuration(final String periodHHMMSSmm){
     }
 
     public static Float convertDurationToSeconds(String duration) {
-        Pattern durationPattern = Pattern.compile("(\\d\\d+):(\\d\\d):(\\d\\d)(:\\d\\d?)?");
+        duration = convertDuration(duration);
```
danielbaker: This is the line which will actually fix the bug!

scoen:
I think that looks good. Cheers Dan

PR #472, opened 2015-11-17
add carrier property to Work

yetti:
Changes to support adding carrier property to Work using a post-commit hook in Banjo.

@danielbaker @MingtaoSun @JonathanShaw @AaronCMuller @scoen @bsingh-nla


AaronCMuller:
I don't really know about the advisability of this, but if this is a requirement then so be it.

PR #474, opened 2015-11-18
fix toJson() to return well formed JSON

yetti:
So, I would like to use toJson() in the CarrierPreCommitHook class in Banjo to find which Work a copy has been deleted from. Unfortunately, the JSON was not well formed at all.

@scoen @AaronCMuller @danielbaker @MingtaoSun


scoen:
much better

PR #476, opened 2015-11-18
ms14: add new method to handle put file with checksum validation

MingtaoSun:
@brycollins @ninhnguyen @scoen 
- add new method to put file with checksum validation


AaronCMuller:
This seems okay.

PR #483, opened 2015-11-25
Schema prep

shuangzhou:
This request includes code refactor to move initialization of lookup data from during amber session creation to amberdb instantiation so that this initialization will only be called once.

### DIFF
--- a/src/amberdb/AmberDb.java
+++ b/src/amberdb/AmberDb.java
@@ -20,14 +20,10 @@ import doss.CorruptBlobStoreException;
 public class AmberDb {
     final private DataSource dataSource;
     final private Path rootPath;
-    static private DBI lookupsDbi;
-    static private Lookups lookups;
     
     public AmberDb(DataSource dataSource, Path rootPath) {
         this.dataSource = dataSource;
         this.rootPath = rootPath;
-        // NLA specific lookup table config
-        lookupsDbi = new DBI(dataSource);
         initLookupData(dataSource);
     }
 
@@ -44,8 +40,10 @@ public class AmberDb {
     }
 
     private void initLookupData(DataSource dataSource) {
+        // NLA specific lookup table config
+        DBI lookupsDbi = new DBI(dataSource);
         LookupsSchema luSchema = lookupsDbi.onDemand(LookupsSchema.class);
-        lookups = lookupsDbi.onDemand(Lookups.class);
+        Lookups lookups = lookupsDbi.onDemand(Lookups.class);
         if (!luSchema.schemaTablesExist()) {
             luSchema.createLookupsSchema();
             List<ListLu> list = lookups.findActiveLookups();
@@ -55,6 +53,9 @@ public class AmberDb {
             luSchema.createCarrierAlgorithmTable();
         }
         lookups.migrate();
+        lookupsDbi.close(luSchema);
+        lookupsDbi.close(lookups);
+        lookupsDbi = null;
     }
     
     static BlobStore openBlobStore(Path root) {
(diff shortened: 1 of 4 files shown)


shuangzhou:
@scoen @terenceingram @brycollins @ato @AaronCMuller could you please have a look?

src/amberdb/AmberDb.java
```diff
@@ -13,10 +20,15 @@
 public class AmberDb {
     final private DataSource dataSource;
     final private Path rootPath;
+    static private DBI lookupsDbi;
+    static private Lookups lookups;
```
ato: Why are these static?

src/amberdb/AmberDb.java
```diff
@@ -13,10 +20,15 @@
...
     final private Path rootPath;
+    static private DBI lookupsDbi;
+    static private Lookups lookups;
     
     public AmberDb(DataSource dataSource, Path rootPath) {
         this.dataSource = dataSource;
         this.rootPath = rootPath;
+        // NLA specific lookup table config
+        lookupsDbi = new DBI(dataSource);
+        initLookupData(dataSource);
```
ato: This is a data race, this clobbers a global variable every time an instance of AmberDb is created.

src/amberdb/AmberSession.java
```diff
@@ -155,30 +149,14 @@ private AmberGraph init(DataSource dataSource, Long sessionId) {
...
-
     public AmberGraph getAmberGraph() {
         return ((OwnedGraph) graph.getBaseGraph()).getAmberGraph();
     }
 
 
     public Lookups getLookups() {
-        return lookupsDbi.onDemand(Lookups.class);
+        // return AmberDb.lookupsDbi().onDemand(Lookups.class);
+        return AmberDb.lookups();
```
ato: Thread-safety issue: a single database connection will be shared by multiple threads. This will likely deadlock.

src/amberdb/AmberDb.java
shuangzhou: @ato, there should only be one instance of AmberDb per application, so AmberDb is currently created once during application startup through AmberDbService singleton, not involved in multiple thread.

ato: Even if that is the case for one particular application it is not true in general and should not be relied upon. For example the unit tests instantiate AmberDb multiple times. I can also forsee several instances when it may be desirable to conntect to two different instances of AmberDb from the one application, such as when doing a data migration. In general this is not an assumption that good library code should ever make.

src/amberdb/AmberDb.java
shuangzhou: @ato, as AmberSession does not have access to the amberdb instance that created it, I'm trying to return the lookupsDbi and lookups through static method like open blob store.  actually are there a way AmberSession can get reference to the data source from the amberdb instance which created it?  If so, I can change the above var to be private final instead.

ato:
Summarising the discussion we had in person:

* Moving the schema migration into AmberDb is good.
* Using a single global Lookups DAO is unnecessary as creating them is relatively cheap and sharing DAOs means sharing database connections which can cause threading issues (particularly when transactions, multi-step temporary-table queries, or high contention are involved)
* The use of statics should be avoided (the correct way to pass objects between the two classes would be do it via the AmberSession constructor, but in this case we don't actually need to do that)

So I think the right approach to this is:

* Create one local Lookups DAO for the schema migration and close it immeidately after schema migration is complete.
* Create one Lookups DAO for each AmberSession in AmberSession.init() and store it as an instance field.

This still means that the session is holding onto two database connections. It would be best to share the same DBI between AmberGraph and Lookups, that might require some more refactoring.

ato:
Actually just realised the Lookups DAO is onDemand, so it won't be holding a database connection. That's not too bad then.

ato:
Looks good to me! :+1:

ato:
@scoen can you give this a quick look too? I'm not super familiar with the AmberDb code and while this looks good to me I might be missing something.

shuangzhou:
@scoen, @ato, @AaronCMuller,I've updated code to include suggested changes, please have another look.

PR #488, opened 2015-12-03
Sc session fix

scoen:
@ato 
Would you have a look ?


src/amberdb/graph/AmberPropertyBatch.java
```diff
@@ -16,6 +16,7 @@
 
     
     void add(Long id, Map<String, Object> properties) {
+        if (this.id.contains(id)) return; // properties already added
```
ato: Does this make the process of building the batch a O(n^2) operation?

scoen: Yeah, it will. 
I'll redo

scoen: yep this check is completely unnecessary

scoen:
@ato 
I've removed that check. Isn't needed.

ato:
> I've removed that check. Isn't needed.

I'm trying to understand the logic for why the check isn't needed. Let me know if I go wrong somewhere.

A vertex can end up in both the "new" and "modified" sets like this:

```java
Work work = session.addWork(); 
// which calls graph.addVertex()
// which I assume calls AmberGraph.addVertes()
// and thus: newVertices.add()

work.setSomething("foo");
// which I assume calls BaseElement.setPropery()
// which calls: graph.elementModListener.elementModified
// and thus: modifiedVertices.add()
```

During suspend() it'll then be saved to the session twice. First as "NEW" in the `for (Vertex v : newVertices) {` loop and then as "MOD" in the `for (Vertex v : modifiedVertices) {` loop.

Then during resume() it'll restored twice as two different AmberVertexWithState objects. The first instance will have state "NEW" and be added to graphVertices and newVertices. The second instance wil have state "MOD" and will overwrite the graphVerices entry and then go into modifiedVertices.

Thus resume() will restores the original state of things with the subtle change that there's two instances of the vertex change. That's surprising but I guess not a problem in practice because the instance in newVertices is never accessed except by suspend() and by newVertices.contains(something). The newVertices.contains() check will succeed because equals() in BaseVertex is just based on the id so it doesn't matter that it's actually a separate object in memory.

scoen:
It's a fair question. 
However when the modified elements are suspended, any elements in both MOD and NEW only go in the NEW set (as it's looped through it removes it from the MOD set). 

eg: (from the batchSuspendEdges method)
`        for (Edge e : newEdges) {`
`            AmberEdge ae = (AmberEdge) e;`
`            modifiedEdges.remove(e);`
`            edges.add(new AmberEdgeWithState(ae, "NEW"));`
`            properties.add((Long) ae.getId(), ae.getProperties());`
`        }`

This means the element only goes in once.

The reason i added the check in this change was that an earlier draft of the changes i was making stored the end vertices of an edge also when it was suspended to try and ensure integrity. I neglected to remove it and it would certainly have impacted performance as you suggested. So it turned out that if the end vertices don't exist in the session it's ok because it reads them from the persistent store.

However, at lunch i realised i need to do something about the scenario when an end vertex is removed from the persistent store before the session is resumed. I'm looking at that now.

scoen:
Just noting: To deal with resuming a stale session (that is: related elements in the persistent datastore have been modified after session suspension but before resumption) i decided that session edges which now don't have 2 current end vertices will not be restored. It shouldn't be a common case. A warning is sent to the log but without major work to Banjo the user is not alerted. I do this rather than throwing some kind of stale session exception so i don't break Banjo. This maintains the graph integrity without completely invalidation the session, but it is a trade-off.

ato:
:+1:

PR #495, opened 2015-12-08
ms1911: add new query to retrieve first and last transactions given a list of work ids

MingtaoSun:
@scoen 
http://ourweb.nla.gov.au/apps/jira/browse/DSCS-1911
I can only think of this straightforward way --- call custom queries to load all transactions upfront, then assign property to each work. I have tested it and it seems to get the job done. Took 1.5 seconds to export 8901 items. 
Related to https://github.com/nla/banjo/pull/1844

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,7 @@
 
   <groupId>au.gov.nla</groupId>
   <artifactId>amberdb</artifactId>
-  <version>1.9.1-SNAPSHOT</version>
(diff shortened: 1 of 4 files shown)


MingtaoSun:
@scoen , Hi Shaun, are you happy to merge? I have more stuff (> 10,000 items CSV export) to check into this branch. And I think by merging this one first would make my next PR easier to review.

PR #496, opened 2015-12-09
Ms1905

MingtaoSun:
@scoen 
- Load copy edge (source copy and file) upfront


src/amberdb/graph/BaseElement.java
```diff
@@ -72,6 +72,8 @@ public void setProperty(String propertyName, Object value) {
     
     
     protected void replaceProperties(Map<String, Object> newProperties) {
-        properties = newProperties;
+        if (newProperties != null){
```
MingtaoSun: @scoen , added this check because there is some corrupt data in Dev that replace the empty Map with a null, causing a NPE.

PR #497, opened 2015-12-09
suspend in batch, rather than all at once

AaronCMuller:
@scoen

### DIFF
--- a/src/amberdb/graph/AmberGraph.java
+++ b/src/amberdb/graph/AmberGraph.java
@@ -250,7 +250,18 @@ public class AmberGraph extends BaseGraph
 
     
     public Long suspend() {
-    
+
+        if (getModifiedElementCount() > BIG_COMMIT_THRESHOLD) {
+            log.warn("Graph to be committed exceeds {} elements. Using big suspend to process.", BIG_COMMIT_THRESHOLD);
+            Long txnId = suspendBig();
+            log.debug("Commence clearing session");
+            dao.clearSession(txnId);
+            log.debug("Finished clearing session");
+
+            clearChangeSets();
+            return txnId;
+        }
+
         // set up batch sql data structures
         Long sessId = newId();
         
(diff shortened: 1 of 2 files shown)


src/amberdb/graph/AmberGraph.java
```diff
@@ -250,7 +250,18 @@ public Vertex addVertex(Object id) {
 
     
     public Long suspend() {
-    
+
+        if (getModifiedElementCount() > BIG_COMMIT_THRESHOLD) {
+            log.warn("Graph to be committed exceeds {} elements. Using big suspend to process.", BIG_COMMIT_THRESHOLD);
+            Long txnId = suspendBig();
```
AaronCMuller: I assume here that I shouldn't be surrounding this in a begin and commit with the dao.

scoen: Yeah, for now your assumption is right. But i do need to revisit that in the near future - i think it's an issue i need to address.

src/amberdb/graph/AmberGraph.java
```diff
@@ -250,7 +250,18 @@ public Vertex addVertex(Object id) {
...
+            log.debug("Commence clearing session");
+            dao.clearSession(txnId);
```
scoen: i think this will delete the session that's just been written to the db

scoen: i think maybe just clear() would clean up memory

AaronCMuller: Oh yeah ... okay.

shuangzhou:
I'll run a test of a large EAD over the change.

brycollins:
Not in production Please.
Last time it caused a 15minute outage for all users.

scoen:
Yes, a test run in devel would definitely be good thing

shuangzhou:
I've tested suspend and commit against creation, reload of 6598 works in both session suspend and session commit, all worked.

PR #501, opened 2016-01-06
retained h2 as upgrade require more testing.

shuangzhou:
H2 v1.3.176 introduced a number of new features that requires more testing, so retained h2 to be at v1.3.172 for now.
@scoen could you please review this change?  thanks.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -87,7 +87,7 @@
     <dependency>
       <groupId>com.h2database</groupId>
       <artifactId>h2</artifactId>
-      <version>1.3.176</version>
+      <version>1.3.172</version>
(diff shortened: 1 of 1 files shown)


scoen:
Hi Shuang, thanks for checking that and providing a solution. I think that's what we need to do until we can address the h2 based tests issue.

brycollins:
This update is highlighting a disconnect between unit tests and runtime.
Nothing uses h2 in runtime, so failures here can be non-indicative of runtime problems.

I agree with merging this pull request for now, until the unit tests can be changed to use a database
that matches runtime.

PR #505, opened 2016-01-08
ms2026: put back deleted getRepImage method

MingtaoSun:
@scoen , I need to quickly merge this as Dev is broken. I was aware that this method was referenced in freemarker template. So it cannot be deleted for now. I need to think about how to fix it in the template. Sorry.


scoen:
No worries Ming - i was rushing you

PR #508, opened 2016-01-11
ms2021: set work order to maxEdgeOrder+1

MingtaoSun:
@AaronCMuller @yetti @shuangzhou 
- Retrieve the max edge order of the children of the specified parent Id
- Related to https://github.com/nla/banjo/pull/1903


yetti:
:+1:

PR #510, opened 2016-02-03
deprecate subtype

AaronCMuller:
@scoen I think this is what we want.

@JonathanShaw


src/amberdb/model/Work.java
```diff
@@ -1857,5 +1878,17 @@ public boolean apply(Work work) {
...
         }
+
+        @Override
+        public Iterable<Work> getChildren(boolean includeSections) {
+            if (includeSections) {
+                List<Vertex> vertices = gremlin().inE(IsPartOf.label).outV().toList();
+                return frameVertices(vertices, Work.class);
+            }
+            else {
+                List<Vertex> vertices = (List<Vertex>) gremlin().inE(IsPartOf.label).outV().hasNot("bibLevel", "Section").toList();
```
scoen: I think Section is a 'type' rather than a 'bibLevel' ?

scoen: sorry - just checked - is also a bibLevel and should work fine

PR #511, opened 2016-02-08
DDG-83: Added new edeposit values to the Constraint lookup table.

sjacob:
For the initial edeposit release, the business area has decided to add 3 new access constraints that can be assigned to Copy.  These values will probably change in future after the edeposit release to production.  I'm not sure if this migration will run automatically or whether manual intervention is required.
@scoen @JonathanShaw @shuangzhou


scoen:
yeah, it'll require manual intervention unfortunately - the migrations for amber are buggered.

PR #512, opened 2016-02-12
Added MovingImageFile class, which is needed for Banjo DSCS-1747

ninhnguyen:
@scoen @JonathanShaw @shuangzhou


shuangzhou:
@ninhnguyen Looks great!

PR #513, opened 2016-02-26
DDG-171: Added support for storing the Preservica type and id.

sjacob:
This change adds support for storing the Preservica type and id.
@terenceingram @scoen @JonathanShaw


scoen:
Looks ok to me

PR #514, opened 2016-03-01
nn material type

ninhnguyen:
Add method to detect material type from mime type.
The next step is to change Banjo to use this method instead of harcoding it everywhere.
@JonathanShaw @scoen @shuangzhou @yetti


shuangzhou:
@ninhnguyen looks good to me.

scoen:
sure

PR #516, opened 2016-03-01
ms2153: add convenience method

MingtaoSun:
Add two convenience methods
@yetti @shuangzhou @josephcurtis @scoen


shuangzhou:
Good idea.

PR #519, opened 2016-03-02
Amberdb update

shuangzhou:


terenceingram:
@scoen @yetti @sjacob @josephcurtis can someone please verify that they are happy with merge?

AaronCMuller:
What on earth ... I don't know how I got this in my timeline, I swear I removed myself from this.

Also, regarding <code>file.getCopy().setFile(file);</code>, given that File is an edge not a property, I don't know why this would need to be set?

@terenceingram

yetti:
@shuangzhou @scoen @sjacob @josephcurtis I agree with @AaronCMuller above. I can understand setting the File _type_, but I don't understand why we need the setFile().

scoen:
Yeah, i think there was a typo - it's being corrected now

PR #521, opened 2016-03-03
update Copy and File impl for reset technical properties method.

shuangzhou:
@scoen thanks for looking at the code with me, I've updated the code with discussed change.
@ninhnguyen @yetti thanks for reviewing the code also.

### DIFF
--- a/src/amberdb/model/File.java
+++ b/src/amberdb/model/File.java
@@ -157,7 +157,7 @@ public interface File extends Node {
     public String getType();
     
     @Property("type")
-    public String setType(String type);
+    public void setType(String type);
 
     /**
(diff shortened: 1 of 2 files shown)


shuangzhou:
@scoen @ninhnguyen could you please review this change?  thanks.

src/amberdb/model/File.java
```diff
@@ -333,27 +333,22 @@ public File resetTechnicalProperties(MaterialType materialType) {
                     this.asVertex().removeProperty(property);
                 }
             }
+            
+            File file;
```
yetti: Good. :+1: I wasn't seeing any advantage to not using the base interface.

yetti:
Looks ok now.

PR #522, opened 2016-03-08
Ms1552

MingtaoSun:
http://ourweb.nla.gov.au/apps/jira/browse/DSCS-1552
related to https://github.com/nla/banjo/pull/2016
@scoen @yetti @josephcurtis @shuangzhou 
- Add a generic work property search query. Note convert(p.value using utf8)=? is case insensitive comparison while p.value = ? is case sensitive comparison.

### DIFF
--- a/src/amberdb/query/PropertyQueryAssembler.java
+++ b/src/amberdb/query/PropertyQueryAssembler.java
@@ -7,6 +7,9 @@ import org.skife.jdbi.v2.util.LongMapper;
 
 import java.util.List;
 
+/**
+ * @see amberdb.query.PropertyQueryAssemblerTest
+ */
 public class PropertyQueryAssembler {
     private static String SQL_TEMPLATE = "select v.id from %s vertex v where v.txn_end=0 %s and v.id=p1.id %s %s";
 
--- a/src/amberdb/query/PropertyQueryAssembler.java
+++ b/src/amberdb/query/PropertyQueryAssembler.java
@@ -1,0 +1,86 @@
+package amberdb.query;
+
+import org.apache.commons.collections.CollectionUtils;
+import org.skife.jdbi.v2.Handle;
+import org.skife.jdbi.v2.Query;
+import org.skife.jdbi.v2.util.LongMapper;
+
+import java.util.List;
+
+public class PropertyQueryAssembler {
+    private static String SQL_TEMPLATE = "select v.id from %s vertex v where v.txn_end=0 %s and v.id=p1.id %s %s";
+
+    private List<WorkProperty> workProperties;
+
+    public PropertyQueryAssembler(List<WorkProperty> workProperties){
+        if (CollectionUtils.isEmpty(workProperties)){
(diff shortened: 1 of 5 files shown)


src/amberdb/query/PropertyQueryAssembler.java
```diff
@@ -0,0 +1,104 @@
...
+            sb.append("and p").append(i+1).append(".txn_end=0 ");
+        }
+        return sb.toString();
+    }
+
+    /**
+     * @return where clause for join Id
+     * e.g. and p1.id=p2.id and p2.id=p3.id
+     */
+    private String whereClauseForJoinId(){
```
scoen: could this just join on the v.id every time ?

scoen:
looks good

src/amberdb/query/PropertyQueryAssembler.java
MingtaoSun: @scoen 

Yeah, your way is simpler. The truth is, when I was playing around with it in MySQLWorkBench, I didn't use vertex table at all. So it started with p1.id = p2.id. When I joined another property table, naturally I added p2.id = p3.id. Then finally  I put vertex table in, I put v.id = p1.id.

When I converted the SQL to Java code, I didn't realise I had the chance to make it simpler until you mentioned it :disappointed:

scoen: it's fine the way it is - was just curious

src/amberdb/query/PropertyQueryAssembler.java
```diff
@@ -0,0 +1,104 @@
...
+    /**
+     * @return pi.value if case sensitive, convert(pi.value using utf8) if case insensitive
+     */
+    private String whereClauseValueColumn(int index, boolean isCaseSensitive){
+        StringBuilder sb = new StringBuilder();
+        if (isCaseSensitive) {
+            sb.append("p").append(index).append(".value");
+        }else{
+            sb.append("convert(p").append(index).append(".value using utf8)");
+        }
```
shuangzhou: @MingtaoSun this might not work for data of type date, might be worth to add in the javadoc that the value is assumed to be of text value.

shuangzhou:
I like the idea of the PropertyQueryAssembler.

src/amberdb/query/WorksQuery.java
```diff
@@ -176,4 +176,26 @@ private static void populateLastTransactionDetails(Work work, Map<Long, AmberTra
...
+     *
+     * Only string(STR) type is catered for now.
+     *
+     * Please be aware of the performance. You shouldn't use it to return a large result set.
+     * For example, you shouldn't use it to return collection = 'nla.pic' and recordSource = 'voyager'
+     *
+     * You should specify at least one search criteria that limits the result to only a few records
+     * For example, return collection = 'nla.pic' and recordSource = 'voyager' and title = 'abcde'
+     * @return
+     */
```
MingtaoSun: @shuangzhou , you are right. It doesn't work for other types than STR. please see the comments I added above.

PR #523, opened 2016-03-10
Introduced sql so modified objects query is now directly through the database, not amber.

josephcurtis:
http://ourweb.nla.gov.au/apps/jira/browse/DSCS-2163

Modified objects query was being run over basically the whole database, and since amber doesn't have any in-built concept of limiting query results, it was ooming.  Split this specific case out into an SQL query so limiting could be introduced.

Please don't delete this branch after merging.  I will likely do some further optimisations.

@scoen @terenceingram @yetti @JonathanShaw @snezanam @MingtaoSun @sjacob


shuangzhou:
Looks good to me.

sjacob:
Looks fine to me.

scoen:
Looks good to me too. Could you just bump the version in the pom before i merge ?

MingtaoSun:
:+1:

yetti:
Looks fine :+1:

AaronCMuller:
Good to see this got fixed.

PR #524, opened 2016-03-10
ms2164: add convenience method

MingtaoSun:
@shuangzhou @yetti 
- add convenience method


shuangzhou:
Looks good to me.

yetti:
There is already a `hasCopyRole(Role)` method. Can you possibly enhance that one instead of making a new method?

MingtaoSun:
@yetti , good idea, I have overloaded the hasCopyRole() method

PR #531, opened 2016-03-16
DDG-294: Added a flag to indicate that onsite access is allowed.

sjacob:
This change adds a new attribute to a work in order to indicate whether onsite access is allowed or not.
@scoen @JonathanShaw @yetti


yetti:
:+1:

PR #533, opened 2016-03-23
Added specific SQL query for trove article indexing.

josephcurtis:
Generic query combined with amber was still too slow for prod.

Translated the logic that was previously in java into sql so we can more efficiently get the data trove needs for indexing.

This monstrous query is very specifically just for this and could be optimised a lot I think.  But it's a lot faster than before.

Related dl-trove-service pull request: https://github.com/nla/dl-trove-service/pull/35

@terenceingram @scoen @yetti @shuangzhou @JonathanShaw @sjacob @MingtaoSun @weijunnla @jrixon

### DIFF
--- a/src/amberdb/query/ObjectsQuery.java
+++ b/src/amberdb/query/ObjectsQuery.java
@@ -105,7 +105,7 @@ public class ObjectsQuery extends AmberQueryBase {
                     "                      AND txn_end <> 0) THEN 'DELETED' ELSE (CASE WHEN (v_count_before = 0) THEN 'NEW' ELSE 'MODIFIED' END) END) AS transition,\n" + 
                     "          'NEW_MODIFIED_DELETED' AS reason\n" + 
                     "   FROM (\n" + 
-                    "           (SELECT p.id,\n" + 
+                    "           (SELECT DISTINCT p.id,\n" + 
                     "                   p.txn_start,\n" + 
                     "                   p.txn_end,\n" + 
                     "                   p.name,\n" + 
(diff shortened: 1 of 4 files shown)


MingtaoSun:
Awesome job, well done :clap:

src/amberdb/query/ObjectsQuery.java
```diff
@@ -24,6 +24,141 @@ public ObjectsQuery(AmberGraph graph) {
         this.vGraph.clear();
     }
 
+    public ModifiedObjectsQueryResponse getArticlesForIndexing(ModifiedObjectsBetweenTransactionsQueryRequest request) {
+        LinkedHashMap<Long, String> modifiedObjects = new LinkedHashMap<Long, String>();
+        LinkedHashMap<Long, String> reasons = new LinkedHashMap<Long, String>();
+
+        try (Handle h = graph.dbi().open()) {
+            h.begin();
+            h.execute("SET @start_transaction = ?", request.getFromTxn());
```
MingtaoSun: So what's the difference between SET @start_transaction, ... and q.bind('start_transaction', ....)

jrixon:
The JDBI docs say that SQL statement caching is done on a per handle basis so it's possible that this (large) query is being re-compiled/re-optimized by the DB every time it is run.
http://jdbi.org/apidocs/org/skife/jdbi/v2/package-summary.html

The DB may be caching the query on the server side so it might not be an issue.

Otherwise, :+1:

src/amberdb/query/ObjectsQuery.java
josephcurtis: SET is assigning an SQL variable as opposed to bind which would insert the value in each place in the query before sending it to the DB.  I think logically there would be pretty much no difference.  It was just the way it is done in the generic query for vertices in amber, and a convenient way to avoid repetition when testing the query in mysql workbench.

josephcurtis:
@jrixon I might have a quick look into whether caching the query would improve performance, but since the query takes about 90 seconds to run in prod anyway, we probably wouldn't see a significant gain.

PR #538, opened 2016-03-31
Commit big suspend

snezanam:
Adding commit after suspendBig method is invoked,
@shuangzhou , @JonathanShaw , @terenceingram, @MingtaoSun   could you please have a look.


shuangzhou:
Looks good.

PR #539, opened 2016-04-01
Commit big suspend

snezanam:
It seems that normal commit does not do the the job, alternating to commitBig when reaching  BIG_COMMIT_TRESHOLD(20000 entries), commitBig calls supendBig  and then does commitSqlWrappedWithRetry. @shuangzhou , @JonathanShaw , @MingtaoSun


JonathanShaw:
I don't think a call to commitBig should be in suspend()  that would commit the session when people are just trying to suspend it, prematurely making the changes public

snezanam:
I am closing this pull request for the time being, need to do another test in devel  with the current amberdb version.

PR #540, opened 2016-04-04
put back begin transaction where it used to be

snezanam:
I am putting the dao.begin() where it used to be(prior the previous change), as there is currently an issue in devel with suspending normal transactions, this is the only obvious change made. Need this looked at urgently as there is an issue in devel.
@JonathanShaw , @shuangzhou , @jrixon , @MingtaoSun , @weijunnla

### DIFF
--- a/src/amberdb/graph/AmberGraph.java
+++ b/src/amberdb/graph/AmberGraph.java
@@ -250,7 +250,7 @@ public class AmberGraph extends BaseGraph
 
     
     public Long suspend() {
-        dao.begin();
+        
         Long sessId = null;
         if (getModifiedElementCount() > BIG_COMMIT_THRESHOLD) {
             log.warn("Graph to be committed exceeds {} elements. Using big suspend to process.",
@@ -271,6 +271,8 @@ public class AmberGraph extends BaseGraph
 
             log.debug("batches -- vertices:{} edges:{} properties:{}",
                     v.id.size(), e.id.size(), p.id.size());
+            
+            dao.begin();
 
             dao.suspendEdges(sessId, e.id, e.txnStart, e.txnEnd, e.vertexOut,
                     e.vertexIn, e.label, e.order, e.state);


jrixon:
:+1:

brycollins:
I've updated devel and UAT's config to match  production.

Please make sure you set these configs  the same, or unexpected results will occur.

AMBER_URL="jdbc:mysql://localhost:3306/amberdb?allowMultiQueries=true&zeroDateTimeBehavior=convertToNull&useUnicode=yes&characterEncoding=UTF-8&relaxAutoCommit=true"

Both env's will need a restart.

weijunnla:
I just got one question, what will happen if calling dao.commit() withhout dao.begin()?

src/amberdb/graph/AmberGraph.java
```diff
@@ -250,7 +250,7 @@ public Vertex addVertex(Object id) {
 
     
     public Long suspend() {
-        dao.begin();
+        
         Long sessId = null;
         if (getModifiedElementCount() > BIG_COMMIT_THRESHOLD) {
```
weijunnla: Why put dao.begin inside else statement? That could cause dao.commit is called wihtout dao.begin. Does that matter?

shuangzhou: @weijunnla Snezana is trying to verify this at the moment.

snezanam: Good point, just doing a test in devel to see what happens when supendBig is invoked, might need to make some changes after the test.

ato: When using JDBI please always use `dao.inTransaction()` and never call begin() or commit() directly unless there is no other choice. If you do call begin() and commit() make sure to wrap them in a try block and handle the rollback case.

This code as written is not safe if an exception is thrown. Similarly, relaxAutoCommit should be unnecessary with correctly written transaction code. i.e. calling commit without begin should be impossible.

ato: For example, something like this:

```java

    public Long suspendBig() {

        // set up batch sql data structures
        final Long sessId = newId();


        dao.inTransaction(new Transaction<Object, AmberDao>() {
            @Override
            public Object inTransaction(AmberDao amberDao, TransactionStatus transactionStatus) throws Exception {
                bigSuspendEdges(sessId);
                bigSuspendVertices(sessId);
                return null;
            }
        });
        log.debug("finished suspend");

        return sessId;
    }

    public Long suspend() {
        
        Long sessId = null;
        if (getModifiedElementCount() > BIG_COMMIT_THRESHOLD) {
            log.warn("Graph to be committed exceeds {} elements. Using big suspend to process.",
                    BIG_COMMIT_THRESHOLD);
            sessId = suspendBig();
            clear();
        } else {

            // set up batch sql data structures
            sessId = newId();

            final AmberEdgeBatch e = new AmberEdgeBatch();
            final AmberVertexBatch v = new AmberVertexBatch();
            final AmberPropertyBatch p = new AmberPropertyBatch();

            batchSuspendEdges(e, p);
            batchSuspendVertices(v, p);

            log.debug("batches -- vertices:{} edges:{} properties:{}",
                    v.id.size(), e.id.size(), p.id.size());

            final Long finalSessId = sessId;

            dao.inTransaction(new Transaction<Object, AmberDao>() {
                @Override
                public Object inTransaction(AmberDao amberDao, TransactionStatus transactionStatus) throws Exception {
                    dao.suspendEdges(finalSessId, e.id, e.txnStart, e.txnEnd, e.vertexOut,
                            e.vertexIn, e.label, e.order, e.state);
                    dao.suspendVertices(finalSessId, v.id, v.txnStart, v.txnEnd, v.state);
                    dao.suspendProperties(finalSessId, p.id, p.name, p.type, p.value);

                    return null;
                }
            });
        }

        return sessId;
    }
```

ato: I realise that inTransaction() is very verbose (it's much nicer in Java 8 using the lambda syntax) but it makes it much harder to make a mistake with the transaction logic - ensuring it's impossible to leave the transaction hanging open by forgetting to rollback or commit.

You can see one of the JDBI developers recommending it here when someone asks about how best to correctly transactions:
https://groups.google.com/forum/#!topic/jdbi/GwpFAGUb5xg

I kind of wish the JDBI site had a best practices page which documented some of this stuff. ;-)

PR #541, opened 2016-04-04
Fix suspend

snezanam:
As Alex suggested ,  put  bigSuspend into dao.inTransaction so the Transaction is committed if execution terminates normally (no exception), but rolled back if an exception is thrown by code. @shuangzhou, @weijunnla , @ato , @MingtaoSun , @jrixon , @JonathanShaw


MingtaoSun:
:+1:

weijunnla:
:+1:

ato:
:+1

PR #543, opened 2016-04-13
DSCS-2124 Re-order by Alias field not sorting in ascending order

jrixon:
Ordering by children by fields causes slightly inconsistent behaviour when handling children that have null in that field. It's better to have a deterministic order for UI ordering.

@yetti @shuangzhou @ninhnguyen @weijunnla @josephcurtis @snezanam


shuangzhou:
Looks good.

PR #544, opened 2016-04-15
ms106: minor string equals change

MingtaoSun:
@shuangzhou @weijunnla @yetti 
- just in case party's name is null although it shouldn't in reality.


shuangzhou:
ok.

PR #546, opened 2016-04-26
update objects query.

shuangzhou:
This request provides slight improvement on query speed for retrieving articles for indexing for the Trove service.  I also noticed that the first subquery for union in this query test for Journal, if the condition can be relaxed to include book and multi-volume books, it'll help with the retrieval speed.


shuangzhou:
@terenceingram @josephcurtis @scoen @jrixon @weijunnla @yetti @snezanam, could you please review this change?  thanks.

shuangzhou:
will close this request for now as @josephcurtis point out there are also the TroveIndexReporterService which requires update to work with this query.

PR #548, opened 2016-04-27
Fixup schema init

jrixon:
@JonathanShaw Jon, this was the problem.


JonathanShaw:
that makes a lot of sense actually

PR #549, opened 2016-04-28
Wz2241 

weijunnla:
Optimize SQL for Duplicate Alias report generation. Since the version has been updated to 1.14.4-SNAPSHOT. Another pull request has been created in banjo to reflect this change. Please let me know if we should not change the version.
@scoen @shuangzhou @JonathanShaw @josephcurtis @sjacob @snezanam


shuangzhou:
@weijunnla looks great!

src/amberdb/query/ObjectsWithPropertyReportQuery.java
```diff
@@ -31,56 +31,29 @@ public ObjectsWithPropertyReportQuery(AmberGraph graph) {
...
+                            + "FROM property p, property cp, property tp \n"
+                            + "WHERE "
+                            + " tp.value in ( :type1, :type3, :type4, :type5) "
+                            + " AND cp.value = :value "
+                            + " AND p.name = :name "
+                            + " AND cp.name = :collection "
+                            + " AND tp.name= :typeName "
+                            + " AND p.txn_end = 0 AND cp.txn_end = 0 AND tp.txn_end = 0 "
+                            + " AND cp.id = p.id AND tp.id = p.id "
+                            + " UNION ALL \n"
```
snezanam: Just wondering why  using UNION ALL  instead of UNION is this case.

weijunnla: UNION has to do additional work to remove the duplicate rows but UNION doesn't. So, UNION ALL has better performance.  In this case there is no duplicate records (because of the distinct in SELECT and other conditions), UNION ALL is a better option.

snezanam: That sounds a better  option then, good to know that  the query is not involving   duplicate records. 
Looks good.

PR #550, opened 2016-05-02
DDG-386-jrixon

jrixon:
DDG-386 Change of collection for legal deposit material

Add Deposit Method field.

@shuangzhou @yetti @ninhnguyen @JonathanShaw @weijunnla @josephcurtis  @sjacob @snezanam @scoen


scoen:
looks good to me

PR #551, opened 2016-05-05
add check to ensure a lookup value is populated.

shuangzhou:
This request provides check to ensure a lookup value is populated during adding or editing of a lookup entry.


shuangzhou:
Looks like banjo app is validating the lookup values before passing the value to amberdb, so I'll close this request.

PR #552, opened 2016-05-10
DDG-386

jrixon:
http://ourweb.nla.gov.au/apps/jira/browse/DDG-386

DepositMethod should be DepositType

@shuangzhou @ninhnguyen @weijunnla @scoen


scoen:
👍 pending travis tests

PR #553, opened 2016-05-12
Sc file size

scoen:
Modify the way getting a file's size works, so that if you use the getFileSize method it will populate the File vertex's fileSize attribute from the blob's reported size if the attribute is null. 

Modified the getSize method (which always tries to get the size from the blob before checking the attribute) so it doesn't call the getFileSize method (as that could call the blob's size method twice). 

I'm not keen on side effects like this change incorporates, but it prevents going to the blob every time you want to know the file's size when calling getFileSize because that can be an expensive/slow activity.

@ato @yetti @josephcurtis @jrixon @sjacob @snezanam @shuangzhou, basically 'at' everyone :-)

### DIFF
--- a/src/amberdb/model/File.java
+++ b/src/amberdb/model/File.java
@@ -220,23 +220,58 @@ public interface File extends Node {
             return getBlobStore().get(getBlobId());
         }
 
+        /**
+         * Tries to return the fileSize attribute of the file vertex first. If this fails,
+         * then goes to the blob to get the file's size. If getting the size from the blob
+         * is successful (or if this file vertex has no blob).
+         *
+         * This method's side-effect is to set's the fileSize attribute to that reported
+         * by the blob, or sets fileSize to 0 if this file has no blob.
+         *
+         * If an IOException occurs, this method returns 0 without a side effect.
+         *
+         * @return the file's size in bytes.
+         */
         @Override
         public long getFileSize() {
             Long fileSize = this.asVertex().getProperty("fileSize");
-            return (fileSize == null)? 0L : fileSize;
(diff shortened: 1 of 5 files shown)


yetti:
What happens if the file gets updated/changed?

src/amberdb/model/File.java
```diff
@@ -251,23 +251,58 @@ private Blob getBlob() throws NoSuchBlobException, IOException {
...
+         */
         @Override
         public long getFileSize() {
             Long fileSize = this.asVertex().getProperty("fileSize");
-            return (fileSize == null)? 0L : fileSize;
+            if (fileSize == null) {
+                try {
+                    fileSize = getBlob().size();
+                    setFileSize(fileSize);
+
```
shuangzhou: @scoen, in amberdb/model/File, we could also rename setBlobId(id) to setBlobIdAndSize(blob), as setBlobId(id) is currently only referenced by 2 methods in the same class: put(...), putWithChecksumValidation(...).  Then in setBlobIdAndSize(), we can call setFileSize(blob.size()).

scoen: good idea - i'll check it out

scoen:
I've addressed Yetrina's issue by making modifications based on Shuang's suggestion and added a couple of tests to check (FileTest.java).

Note: i cleaned up a couple of tests along the way so they extend the AbstractDatabaseIntegrationTest class to make them faster (WorkTest.java and CopyTest.java). These are unrelated to this change except that i saw them and thought i'd do a quick tidy.

@yetti or anyone ?

yetti:
Looks good 👍

PR #555, opened 2016-05-18
DDG-411: Fixed the findModelByValueInJsonList so that query values can contain special regular expression characters

sjacob:
Fixed the findModelByValueInJsonList so that query values can contain the characters (, ), [, ], ., {, }.
@scoen @shuangzhou


jrixon:
👍

PR #557, opened 2016-05-30
DDG-413: Added 'Article' as a new subunit type.

sjacob:
This change adds 'Article' as a new subunit type.
@scoen @jrixon @weijunnla @shuangzhou


jrixon:
👍

PR #567, opened 2016-07-19
refactored models and mapper to use reflection

yetti:
@jrixon 

I've refactored the models to use `@Entity` and `@Column` annotations, so we're not using a custom annotation anymore. I also refactored the mapper factory and created a new generic mapper `AmberDbResultSetMapper`.

`AmberDbResultSetMapper` maps the POJO properties by matching the column name with what's returned on the `ResultSetMetaData`, then it uses BeanUtils to convert and set the value on the POJO.

I've not tested this yet, since I don't have a working copy of the database. Just thought I'd put this out here for your feedback. The introspection of private fields is a bit icky, but saves time building mappers for multiple POJOs.


shuangzhou:
👍

PR #569, opened 2016-07-22
DSCS-345 puts the public access condition check back into the rules f…

m-r-c:
…or determining whether a work can be used as a representative image as it made the image available in tarkine

Original pull request: https://github.com/nla/amberdb/pull/565

@scoen


scoen:
👍

PR #570, opened 2016-07-28
Join tables and other horrible things

yetti:
Mainly implemented join tables and query services.

* `AmberDbResultMapper` was updated to take into account that inheritance is 2 levels now, since some classes have been changed to extend from `File` and `Work`.
* Join tables for the following relationships were created:
  * Copy <=> File (copy_file)
  * Parent <=> Child (parent_child)
  * Work <=> DeliveryWork (work_deliverywork)
  * Work <=> Section (work_section)
* Models and query services were created for above tables in the `relation` package.
* `AmberSession` modified to manage an instance of new DAOs. _Possibly not the best thing to do this on `init()`. Could probably do this on `getXxxDao()` instead._
* Abstract base class `AmberPreCommitHook` no longer deals with Lists of vertices, edges and properties, but subclasses of `AmberModel`. NEEDS LOTS OF REGRESSION TESTING!! Especially `AggregateDurationDeliveryWorkCommitHook` because it relied heavily on `Edges`.
  * Pre-commit hooks were updated to use updated and relocated `AmberPreCommitHook`.
* Renamed `TestUtils` to `AmberDbTestUtils` so JUnit4 wouldn't pick it up and try to run it as a test.

**TODO**
* Fill out the missing CRUD methods in some of the DAOs.
* Clean up TODOs in `Copy` and `Work` models since many of them are done by the query services now.
* Finish off TODOs in `AmberSession`


jrixon:
👍 👍 👍 👍 👍 👍 👍 👍

PR #573, opened 2016-08-08
All vertices and edges now persist to the flat tables.

jrixon:
[New pull request to merge into amberv2 not master!]

@yetti 
Hi Yetrina,

If you look at AmberGraph.vertexToTableMap and AmberGraph.edgeToTableMap you'll see that we now have a mapping between vertex types and tables and also edge types and tables, as discussed on Friday (or Thursday?).

work, eadwork, page, copy and section are all stored in the work table.
file, imagefile, movingimagefile, and soundfile are all stored in the file table.
description, cameradata, geocoding and iptc are all stored in the description table
party and tag have their own tables.

The edges are all currently stored in one table called flatedge. TODO: think of a better name.

The acknowledge edge should have its own table with appropriate fields but I haven't done that yet.

I have created all the needed tables in the amberdb2 database. Doing so has meant losing the history in the history tables. That will be my first priority next.

Actually, my first priority next will probably be to get the unit tests working again.

I have clicked around and tested a few things. I can create/modify/delete works and copies and children and it seems to work. I can see the data going in to the tables.

Talk to you tomorrow to have a sync up with what you're doing.

### DIFF
--- a/src/amberdb/graph/AmberEdge.java
+++ b/src/amberdb/graph/AmberEdge.java
@@ -132,4 +132,15 @@ public class AmberEdge extends BaseEdge implements Comparable {
 		return txnEnd;
 	}
 
+	public Long getInId() {
+		return inId;
+	}
+
+	public Long getOutId() {
+		return outId;
+	}
+
+	public Integer getOrder() {
+		return order;
+	}
 }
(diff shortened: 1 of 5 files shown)


yetti:
@jrixon Looks good, but can you change `Partys` to `Parties`? 

Also, don't forget to update the methods that create the tables. They get called by the tests to create them in the H2 DB.

jrixon:
@yetti No more Partys :D

yetti:
👍

PR #579, opened 2016-09-01
DDG-857: Enabled data stored in IPTC objects to be included in work reports.

sjacob:
When banjo runs a report, it sets the amber session to local mode (ie: cache only mode).  This means that only work data that is cached in the session can be included in the report.  This change causes the IPTC object to be cached in the session.
@scoen @m-r-c @wibing @shuangzhou


nlabuilder:
Build finished.

shuangzhou:
👍

PR #580, opened 2016-09-05
Add IPTC data to work

scoen:
@sjacob
Hi Simon, 
Just thought i'd pull in the change you made to this method 3 days ago before amberdb became the new rdb version. Is that ok ?


sjacob:
@scoen  That's okay with me.

PR #585, opened 2016-09-16
Change "captions" to a JSON field.

jrixon:
@scoen @yetti @shuangzhou @ninhnguyen @m-r-c @josephcurtis @weijunnla


src/main/java/amberdb/model/Section.java
```diff
@@ -70,12 +97,47 @@
...
+	    static ObjectMapper mapper = new ObjectMapper();
+	    
         public int countExistsOns() {
             return (existsOns() == null)? 0 : existsOns().size();
         }
 	    
         private List<Edge> existsOns() {
             return (gremlin().outE(ExistsOn.label) == null)? null: gremlin().outE(ExistsOn.label).toList();
         }
+
```
yetti: Aren't these three JSON de-/serialisation methods already existing on `Work`?

jrixon: Yep, I copied them from Work.Impl. I didn't want to change the hierarchy of Section.Impl and I thought these methods are pretty small.

However, you are correct - I shouldn't be duplicating code. I'll fix it up.

jrixon: Done.

PR #592, opened 2016-09-22
DSCS-2363 Update AmberSession.java to use AmberDB v2 tables

jrixon:
@scoen @yetti @ninhnguyen @shuangzhou @weijunnla @m-r-c


yetti:
👍

PR #594, opened 2016-09-28
update VersionQuery to use Amber V2 tables

yetti:
@scoen @jrixon @m-r-c @weijunnla @shuangzhou @ninhnguyen 

Had to make a couple `private` static fields in `AmberVertexMapper` and `AmberQueryBase` public so I could reuse it instead of duplicating the same code.

I'm also not sure if `VersionQueryTest` ever tested correctly, given that we've created a book with 15 pages, then we're deleting a page from the book on line 106. Anyhow, I've updated the test.

### DIFF
--- a/src/main/java/amberdb/graph/AmberVertexMapper.java
+++ b/src/main/java/amberdb/graph/AmberVertexMapper.java
@@ -26,6 +26,11 @@ public class AmberVertexMapper implements ResultSetMapper<AmberVertex>  {
     	skipProps.add("txn_start");
     	skipProps.add("txn_end");
     	skipProps.add("state");
+		skipProps.add("step");
+		skipProps.add("vid");
+		skipProps.add("eid");
+		skipProps.add("label");
+		skipProps.add("edge_order");
     }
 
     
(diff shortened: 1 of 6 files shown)


yetti:
Will close this for now until I fix the other test that's breaking. :(

yetti:
Let's try this again...

@scoen @jrixon @m-r-c @weijunnla @shuangzhou @ninhnguyen please have a look

nlabuilder:
Build finished.

src/main/java/amberdb/graph/AmberVertexMapper.java
```diff
@@ -20,12 +20,17 @@
     
     private AmberGraph graph;
     
-    private static Set<String> skipProps = new HashSet<>();
+    public static Set<String> skipProps = new HashSet<>();
     static {
     	skipProps.add("id");
     	skipProps.add("txn_start");
     	skipProps.add("txn_end");
     	skipProps.add("state");
```
yetti: @jrixon Needed to add these properties to the set of properties to skip. The mapper was adding the temporary table fields to the Vertex properties after saving SoundFiles and causing the `AmberDao::suspendIntoFlatVertexTable()` method to fail with an SQL error.

nlabuilder:
Build finished.

src/main/java/amberdb/graph/AmberVertexMapper.java
jrixon: Yep, looks fair enough.

PR #609, opened 2016-10-19
add "Access Agreement" field to work

weijunnla:
@scoen @yetti  Could you please review?


scoen:
It looks good to me.
We might have to manually update the lookups in prod/uat when the time comes because the old amberdb migrations are all messed ie: they probably won't run.

weijunnla:
@scoen I see. I'll put a memo in JIRA.

PR #619, opened 2016-10-26
Sz result iterator

shuangzhou:
Another good case for ResultIterator as some of the finding aid work set could grow very big.
see ref: http://jdbi.org/sql_object_api_queries/


jrixon:
Looks good - it will at least avoid some duplication in memory. If it doesn't resolve the issue then we can look at other things to do as well.

PR #622, opened 2016-11-01
Sc add indexes

scoen:
@jrixon 
Would it be ok to merge this to production branch also? 

It incorporates your other index tweeks (so prod migration script will be the same as uat), but was unsure if you wanted the prod version to remain as is for any reason. 

If ok i can redo the release tags to incorporate this change into the 2.0.0 release version of amberdb.

Otherwise, happy to just run the uat version of the script for this upcoming prod migration.


shuangzhou:
👍

PR #633, opened 2016-11-16
Fix wrong work field name in sql

weijunnla:
@m-r-c @yetti @shuangzhou


nlabuilder:
Build finished.

src/main/java/amberdb/graph/dao/AmberDao.java
```diff
@@ -404,7 +404,7 @@
 			+" commentsInternal TEXT,"
 			+" commercialStatus VARCHAR(63),"
 		    +" commercialStatusReasonForChange VARCHAR(63),"
-			+" commercialStatusReasonForChangeSum VARCHAR(63),"
+			+" commercialStatusReasonForChangeSum VARCHAR(126),"
```
shuangzhou: Should this field have a larger size maybe 255?

weijunnla: Yes

weijunnla:
@yetti These fields store text as the template shown: Lookup value of Primary field (e.g. Access Access lookup value) + Lookup value of Reason for Change  + (change by username on dd/MM/yyyyy). I think VARCHAR should be ok.

PR #641, opened 2016-11-27
correct the lookup code for commercial status.

shuangzhou:
Update the lookup code from Non-commerical to Noncommerc for the Non-commercial value.


nlabuilder:
Build finished.

scoen:
nice catch 👍

PR #643, opened 2016-11-28
adding mysql replace for dup entry work around.

shuangzhou:
This provides additional guarantee for not inserting dup entry in the persistence data.


shuangzhou:
@scoen @jrixon could you please review?  Thanks.

scoen:
I think that looks fair enough.
Note: will likely not work on h2

jrixon:
@scoen we don't have the primary keys in the h2 db so we don't need to make the same change in the h2 dao

jrixon:
@shuangzhou 
Looks good.

shuangzhou:
Thanks both :-)

scoen:
@jrixon @shuangzhou 
Yeah, it's a minor thing - not significant here. I'm just not certain h2 understands the `REPLACE` statement as valid sql.

PR #644, opened 2016-11-28
DSCS-2447

jrixon:
Remove some fields from the Acknowledge object before persisting because they shouldn't be there.


jrixon:
@scoen

scoen:
👍

PR #646, opened 2016-11-29
Change non commercial lookup code from NonCommerc to Noncommerc

weijunnla:
Final decision is made to change non commercial lookup code from NonCommerc to Noncommerc

@scoen @yetti @sjacob


shuangzhou:
👍

PR #650, opened 2016-12-07
Sz dscs 2425

shuangzhou:
This request logs and reports records with data too big for amberdb fields in order to help with locating and amended these values in job processing and work records creation.

### DIFF
--- a/src/main/java/amberdb/graph/AmberGraph.java
+++ b/src/main/java/amberdb/graph/AmberGraph.java
@@ -349,6 +349,7 @@ public class AmberGraph extends BaseGraph
             } catch (org.skife.jdbi.v2.exceptions.UnableToExecuteStatementException err) {
                 try {
                     String largestRecord = logLargestRecordInSession(sessId, p);
+                    largestRecord = (largestRecord == null || largestRecord.isEmpty())? "" : "Failed as record is too big: " + largestRecord;
                     throw new org.skife.jdbi.v2.exceptions.UnableToExecuteStatementException(largestRecord, err);
                 } catch (JsonProcessingException pe) {
                     throw err;
--- a/src/main/java/amberdb/graph/AmberGraph.java
+++ b/src/main/java/amberdb/graph/AmberGraph.java
@@ -9,6 +9,7 @@ import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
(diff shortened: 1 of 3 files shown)


shuangzhou:
@scoen @jrixon could you please review? Thanks.

jrixon:
I think this could result in some misleading error messages in the logs. It's not certain that the UnableToExecuteStatementException exception is caused by values being too long - the exception could have been caused by something else entirely (table locking problems, for example).

shuangzhou:
@jrixon I can also check for the wrapped exception and verify the message contains Data too long for column?

scoen:
i think also, we probably don't want the entire too-big entry going into the logs. Maybe just the start of it so it can be identified (maybe include id of work/copy)? Very long lines in the logs makes some shells/viewers go a bit funny when trying to read them.

shuangzhou:
@scoen I'll update to have the first 500 characters.

nlabuilder:
Build finished.

src/main/java/amberdb/graph/AmberGraph.java
```diff
@@ -358,6 +371,35 @@ public Long suspend() {
...
+        }
+        log.error("Failed to suspend due to properties in session: {}", sessId);
+        if (recordToLog != null) {
+            for (String field : recordToLog.keySet()) {
+                log.error("Values of largest record in session {} : {} : {}", sessId, field, recordToLog.get(field));
+            }
+        }
+        return (new ObjectMapper()).writeValueAsString(recordToLog);
+    }
+
```
scoen: Not sure the linked maps really add much value to the routine. Can you do something like: 

```
// loop through all values
int indexOfLargest = -1;
int largestSize = 0;
for (int i = 0; i < p.value.size; i++) {
    if (p.value.get(i).length > largestSize) {
        indexOfLargest = i;
        largestSize = p.value.get(i).length;
    }
    // now use indexOfLargest as index in the other AmberPropertyBatch arrays to extract name etc as required
    // eg: long idOfLargest = p.id.get(indexOfLargest);
}

```

scoen: sorry , pull the final comments out of the loop
```// loop through all values
int indexOfLargest = -1;
int largestSize = 0;
for (int i = 0; i < p.value.size; i++) {
    if (p.value.get(i).length > largestSize) {
        indexOfLargest = i;
        largestSize = p.value.get(i).length;
    }
}
// now use indexOfLargest as index in the other AmberPropertyBatch arrays to extract name etc as required
// eg: long idOfLargest = p.id.get(indexOfLargest);

nlabuilder:
Build finished.

nlabuilder:
Build finished.

nlabuilder:
Build finished.

nlabuilder:
Build finished.

nlabuilder:
Build finished.

scoen:
FYI interested parties :-) Swaps the order that sessions are suspended in, to be flat tables before old graph tables. This is so now, when we try to suspend a value that's too big for a given column, the exception that gets thrown tells us which column has the problem rather than just saying some property value is too big.

PR #672, opened 2017-02-22
A work around to get children in order

ninhnguyen:
@scoen @sjacob @m-r-c @shuangzhou 
This is a work around to get children of a work in order.
For some reason, the edge_orders of children of some works are the same, eg.
select * from flatedge where v_in = 38235327 order by edge_order;
The edge_order is 1 for all children.
The DISTINCT in the query causes the query to result in a list of arbitrary order children.
The work around is to add the flatedge id as the second sort param, with the assumption that an edge with a lower id is added before one with a higher id.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,7 @@
 
     <groupId>au.gov.nla</groupId>
     <artifactId>amberdb</artifactId>
-    <version>2.3.0-SNAPSHOT</version>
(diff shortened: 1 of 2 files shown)


shuangzhou:
👍

weijunnla:
@ninhnguyen  I think I just recently fix the order issue but that fix may not be available in prod.

PR #675, opened 2017-02-27
Silence doss migrations during mvn test

scoen:
Complains about multiple logback.xml configs, but does stop the logging spam during `mvn test`


shuangzhou:
👍

PR #678, opened 2017-03-14
refactor from findbugs

shuangzhou:
This request check the state before getting the state name as there's possibility of state being null.


nlabuilder:
Build finished.

PR #680, opened 2017-03-31
DSCS-2541 A big Spreadsheet ingest fail

potterynla:
@scoen 

Can you please have a look this change. this change is used to improve alias pi querying performance to fix big spreadsheet importing  timeout issue.

### DIFF
--- a/src/main/java/amberdb/graph/AmberGraph.java
+++ b/src/main/java/amberdb/graph/AmberGraph.java
@@ -1030,4 +1030,20 @@ public class AmberGraph extends BaseGraph
         	}
         }
 	}
+	
+    public Iterable<Vertex> getVerticesByAliasValue(String key, String value) {
+        if (!localMode) {
+            AmberVertexQuery avq = new AmberVertexQuery(this); 
+            avq.executeAliasSearch(value);
+        }
+        
+        List<Vertex> vertices = new ArrayList<>();
+        for (Vertex vertex : graphVertices.values()) {
+            String s = vertex.getProperty(key);
+            if (s != null && s.contains("\""+value+"\"")) {
+                vertices.add(vertex);
+            }
+        }
+        return vertices;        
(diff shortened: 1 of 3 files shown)


nlabuilder:
Build finished.

src/main/java/amberdb/graph/AmberGraph.java
```diff
@@ -1030,4 +1030,20 @@ private void suspendIntoFlatEdgeTables(Long sessId, Map<String, Set<AmberEdge>>
...
         }
 	}
+	
+    public Iterable<Vertex> getVerticesByAliasValue(String key, String value) {
+        if (!localMode) {
+            AmberVertexQuery avq = new AmberVertexQuery(this); 
+            avq.executeAliasSearch(value);
+        }
+        
+        List<Vertex> vertices = new ArrayList<>();
```
scoen: nice validation check

src/main/java/amberdb/graph/AmberVertexQuery.java
```diff
@@ -188,4 +188,12 @@ protected String generateQuery() {
         }
     }
 
+	
+    public List<Vertex> executeAliasSearch(String value) {
+        String sql = "select * from node left join alternate_pi on alternate_pi.obj_id = node.id where alternate_pi.legacy_id = :value";
+        try (Handle h = graph.dbi().open()) {
+            Query<Map<String, Object>> q = h.begin().createQuery(sql).bind("value", "%" + value + "%");
```
scoen: i think it's preferable to do an exact match here rather than over-match. 
The Oral History collection has aliases that are substrings of other aliases (i think). Also, the validation in `getVerticesByAliasValue` would exclude over-matches anyway so it doesn't gain anything.

nlabuilder:
Build finished.

scoen:
👍

PR #682, opened 2017-06-15
DSCS-2582  Additional functionality for tags

potterynla:
@scoen @yetti @ninhnguyen 

Please have a look this.


scoen:
@yetti @ninhnguyen 
Some background for Yetrina and Ninh:
This query gets used by a method that sets the amberSession into localMode. That means it only sees what's in memory for the session - it doesn't query the database to find objects. The change ensures that Tags related to the Works or Copies in question are also loaded into memory so they can be found during the localMode. 

The reason for setting localMode is to prevent large numbers of database accesses when looping over large numbers of Works or Copies.

Hope that helps (but perhaps as clear as mud).

PR #693, opened 2017-11-27
Fixed children sort bug (DL-33)

ninhnguyen:
@sjacob @scoen 
Bug with ambiguous sortBy column (eg. accessConditions, alias, etc.) as it exists in both node and work.


nlabuilder:
Build finished.

PR #699, opened 2018-04-06
DL-54: Quote white spaces so that they aren't ignored in the sort

ninhnguyen:
NaturalSort uses RuleBasedCollator (https://docs.oracle.com/javase/8/docs/api/index.html?java/text/RuleBasedCollator.html) which ignores white spaces when doing the sorting. The work around is to turn white spaces into a single quoted space to make the sort behaves correctly.
@m-r-c @adityaburra @sjacob


nlabuilder:
Build finished.

sjacob:
👍

PR #700, opened 2018-05-09
DL-60: Added support for calculation duration of Moving Image

ninhnguyen:
The calculation of duration only applies to Sound material type.  Extended it to include Moving Image (DL-60).
@nlahuwgeddes @adityaburra @m-r-c @sjacob


nlahuwgeddes:
👍

nlabuilder:
Build finished.

